PR #10, opened 2018-05-02
Testing

zshaheen:
* Added testing script `tests/test.py`
* Restructured package to make it installable.
* Added support for `hsi chgrp`.


zstash/utils.py
```diff
@@ -60,18 +58,20 @@ def addfiles(cur, con, itar, files):
...
-            failures.append(file[0])
+            traceback.print_exc()
+            logging.error('Archiving %s' % (file))
+            failures.append(file)
 
         # Close tar archive if current file is the last one or adding one more
         # would push us over the limit.
-        if (i == nfiles-1 or tarsize+files[i+1][1] > config.maxsize):
+        next_file_size = tar.gettarinfo(file).size
+        if (i == nfiles-1 or tarsize+next_file_size > config.maxsize):
```
forsyth2: As far as I can tell, this is incorrect. Previously, we were actually comparing against the size of the _next_ file (`file = files[i]` where `file[1] = os.path.getsize(file[0])`. After this change, we simply run `tarsize+next_file_size`, but `next_file_size` is the size of the file we _already_ added with `addfile`

PR #12, opened 2018-11-16
Check

zshaheen:
* Added functionality for `zstash check`, which checks the checksum of all files in a given HPSS archive against the checksum calculated before upload which itself is stored in `zstash/index.db`.


zshaheen:
The updated documentation is here: https://e3sm-project.github.io/zstash/docs/html/quickstart.html#check

PR #18, opened 2019-02-05
List

zshaheen:
* Added `zstash ls` functionality to list the contents of an HPSS archive.
* Support for `-l` argument as well, for more information.
* I'll add documentation on this before we release another version with this support. I promise.


zshaheen:
- [ ] Add tests for this.

Issue #15, opened 2019-01-31
Implement verbosity command line option


golaz:
This should be easy. Implement a new command line option to control verbosity of zstash (i.e. select between INFO, DEBUG, etc.).

zshaheen:
When doing so, I might add a new level underneath `DEBUG` for some of the parallel debugging functionality. Mainly, it would be for the below statement in `extract.py` which prints what each worker is doing.

```python
if multiprocess_worker:
    print('{} is {} {} from {}'.format(multiprocess_worker, cmd, file[1], file[5]))
```

golaz:
Done in 9814be8693f6ab2388601d42634f0f9f2d545677

PR #35, opened 2019-10-04
Verbosity

golaz:
Implement verbosity option. By default, zstash will now only print messages with level INFO and above. DEBUG messages can be printed using the new -v / --verbose command line option.

When using multiple workers for extract/check, verbosity is set to DEBUG regardless of -v option. There is no impact on functionality. Trying to lower it would cause zstash to hang, which appears to be related to the custom logger implemented for parallel processing. 

Tests have been updated and they all pass.


golaz:
Addresses https://github.com/E3SM-Project/zstash/issues/15

Issue #21, opened 2019-02-12
Add option to work without HPSS


golaz:
E3SM dedicated machines (anvil, compy) do not have HPSS storage attached to them. It is currently up to individual users to decide on the best way to free disk space when needed. This is inefficient and could lead to unfortunate data loss.

One remedy would be to implement an option for zstash to work on systems without HPSS. This could take two forms:

1. Only create local tar archives in the zstash subdirectory. The user could then transfer tar and index.db files for offline storage.
2. Implement a remote HPSS option using globus.

The command line usage might look something like:

`zstash create --hpss=none ...` for local only

`zstash create --hpss=globus:nersc-hpss/... ...` for remote HPSS using globus.

In terms if implementation, this may be reasonably straightforward. The bulk of the changes would consist of generalizing

https://github.com/E3SM-Project/zstash/blob/master/zstash/hpss.py

to handle different backends.

forsyth2:
Step 1 ("Only create local tar archives in the zstash subdirectory. The user could then transfer tar and index.db files for offline storage.") is resolved in #42. Step 2 ("Implement a remote HPSS option using globus.") still needs to be implemented.

golaz:
Let's close this for now. We can open a new issue if we decide to proceed integrating Globus transfer into zstash.

PR #39, opened 2020-01-22
Refactor hpss code and tests

forsyth2:
In zstash/hpss.py:
- Factor out common code into helper functions.
- Add option to not use HPSS (thus completing step 1 of 2 in #21)

In tests/test.py:
- Refactor test code to use `unittest`.
- Add test for when `hpss='none'`.

### DIFF
--- a/tests/test.py
+++ b/tests/test.py
@@ -1,14 +1,11 @@
 """
-Run the test suite like:
-    python test.py
-You'll get a statement printed
-if all of the tests pass.
+Run the test suite with `python test.py`
 """
 
 import os
-import sys
-import subprocess
 import shutil
+import subprocess
+import unittest
 from collections import Counter
 
 def write_file(name, contents):
@@ -16,7 +13,8 @@ def write_file(name, contents):
     Write contents to a file named name.
     """
     with open(name, 'w') as f:
-        f.write(contents)   
+        f.write(contents)
+
 
 def run_cmd(cmd):
     """
@@ -39,29 +37,10 @@ def run_cmd(cmd):
     print(err)
     return output, err
 
-def str_not_in(output, msg):
-    """
-    If the msg is not in the output string, then everything is fine.
-    """
-    if msg in output:
-        print('*'*40)
-        print('This was not supposed to be found: {}'.format(msg))
-        print('*'*40)
-        stop()
 
-def str_in(output, msg):
+def cleanup(hpss_path):
     """
-    If the msg is in the output string, then the everything is fine.
-    """
-    if not msg in output:
-        print('*'*40)
-        print('This was supposed to be found, but was not: {}'.format(msg))
-        print('*'*40)
-        stop()
-
-def cleanup():
-    """
-    After this script is ran, remove all created files, even those on the HPSS repo.
+    After the script has failed/run, remove all created files, even those on the HPSS repo.
     """
     print('Removing test files, both locally and at the HPSS repo')
     if os.path.exists('zstash_test'):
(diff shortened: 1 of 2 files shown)


zstash/hpss_new.py
```diff
@@ -0,0 +1,74 @@
...
+    status = completed_process.return_code
+    if status != 0:
+        logger.error(error_str)
+        logger.debug('stdout:\n%s', completed_process.stdout)
+        logger.debug('stderr:\n%s', completed_process.stderr)
+        raise Exception
+
+
+def hpss_transfer(hpss, file_path, put, keep=None):
+    if hpss == 'none':
```
golaz: Should we make 'none' cas insensitive, just in case someone uses Python capitalization None.

zstash/hpss_new.py
```diff
@@ -0,0 +1,74 @@
...
+
+
+def hpss_get(hpss, file):
+    """
+    Get a file from the HPSS archive.
+    """
+    hpss_transfer(hpss, file, False)
+
+
+def hpss_chgrp(hpss, group, recurse=False):
```
golaz: Don't remember where this is called from on top of my head. Do we need to do anything here?

forsyth2: It's used in chgrp.py (https://github.com/E3SM-Project/zstash/blob/master/zstash/chgrp.py#L23). Yeah I should probably add the "none" check to this function too.

tests/test.py
```diff
@@ -70,448 +49,594 @@ def cleanup():
...
+                f_name = '/'.join(f_name[1:])
+                f_hash = l[0]
+                local_dict[f_name] = f_hash
+        print('filename|HPSS hash|local file hash')
+        for k in local_dict:
+            print('{}|{}|{}'.format(k, hpss_dict[k], local_dict[k]))
+
+        cleanup(hpss_path)
+
+    def testZstashWithoutHPSS(self):
```
forsyth2: Is this test case sufficient for testing the `hpss='none'` case? I just 1) removed failing lines and 2) the test segments that evidently couldn't be tested without HPSS.

forsyth2:
@golaz Ready for review. Both test cases pass on Cori. I couldn't seem to get my environment set up properly to test on Compy. I imagine I'll need to update the with-hpss test to check if we even have HPSS.

golaz:
@forsyth2, thanks for working on this; it's a good start. I performed some tests on cori today and identified one major issue and a few minor ones that need to be addressed before merging.

**Major issue**

`zstash check hpss=none` will delete the tar files. This would be a disaster. Users archiving their data, checking that it has been archived properly, and then realizing that it is gone! I think it has to do with the internal flag 'keep' being 'False'. Looks like changing the code at the hpss level only as I suggested is not sufficient, we need changes higher up. To reproduce:

```
mkdir test_files
vi file1.txt
vi file2.txt
cd ..
zstash create --hpss=none test_files
ls -1 test_files/zstash/
000000.tar
index.db
cd test_files
rm file1.txt file2.txt
zstash check --hpss=none
INFO: Opening tar archive zstash/000000.tar
INFO: Checking file1.txt
INFO: Checking file2.txt
INFO: No failures detected when checking the files.
ls -1 zstash
index.db
#--> tar file has been removed!!!
```

For extra safety when using hpss=none, we should make the tar files read-only when closing them.

**Minor issues**

1. Implement a test to catch the condition above. 
2. The tests without HPSS still performs many hsi operations and therefore won't be able to run on compy/anvil. We need to remove all hsi operations from the tests without HPSS.
3. Implement an option to only run tests without HPSS for testing on compy/anvil.
4. There is an error message in the log when testing without HPSS:
```
+ find zstash_test_backup -regex .*\.txt.* -exec md5sum {} + 
find: âzstash_test_backupâ: No such file or directory
```
5. Figure out what is wrong with 'chgrp' and re-activate the specific test.
6. Increment version number.

We also need to resolve issues that you encountered when trying to install on Compy. Users should be able to install and use zstash there.

Let me know if you need any help or clarification. We can look at the issues above together if you'd like.

forsyth2:
@golaz Thanks for the feedback. If properly implementing the `none` option (and tests) is going to be a bit more involved, perhaps I should take out those changes for now, get `chgrp` passing, merge this PR, make a new PR for implementing `none`, and increment the version after both PRs merge?

golaz:
@forsyth2, I agree it's a good idea to break the work into two separate PRs to keep their scopes clean.

forsyth2:
Closing in favor of #40 (to refactor) and a future pull request (to add `hpss='none'` option).

PR #40, opened 2020-01-28
Refactor HPSS code and tests

forsyth2:
This pull request partially replaces #39, to refactor HPSS code and tests.

In zstash/chgrp.py:
- Add import statement to get test to pass.

In zstash/hpss.py:
- Factor out common code into helper functions.

In tests/test.py:
- Refactor test code to use unittest.

### DIFF
--- a/tests/test.py
+++ b/tests/test.py
@@ -1,22 +1,21 @@
 """
-Run the test suite like:
-    python test.py
-You'll get a statement printed
-if all of the tests pass.
+Run the test suite with `python test.py`
 """
 
 import os
-import sys
-import subprocess
 import shutil
+import subprocess
+import unittest
 from collections import Counter
+
 
 def write_file(name, contents):
     """
     Write contents to a file named name.
     """
     with open(name, 'w') as f:
-        f.write(contents)   
+        f.write(contents)
+
 
 def run_cmd(cmd):
     """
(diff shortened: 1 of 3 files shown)


forsyth2:
@golaz Ready for review. Once this merges, I'll work on the `hpss='none'` pull request.

forsyth2:
@golaz Thanks for reviewing. 

Wouldn't https://github.com/E3SM-Project/zstash/pull/40/files#diff-b131fe2646b3e4980d06f90692dc2401R470 record any mismatches though?

Interesting that your output is out of order. When I run the test, it ends with "Ran 1 test [...] OK". Looks like I can just add a `,flush=True` to `print` calls. I can add that to https://github.com/E3SM-Project/zstash/pull/40/files#diff-b131fe2646b3e4980d06f90692dc2401R38 to have it flush every time a command is run. Would that be good? Or do we only want it flushed at the very end?

golaz:
@forsyth2, I agree, mismatches should be caught somewhere higher up in the test. I don't know if @zshaheen had the intention to add one more level of verification by separately computing md5sum and checking them against the database. We can leave this as is for now.

I've noticed before that the output order (and flushing) can change depending on how a Python script is invoked. I was using:

`python test.py 2>&1 | tee test.log`

If adding `flush=True` to the `print` call takes care of it, that's sufficient.

forsyth2:
@golaz Ok, I'll leave the checking as it is now. And yes, it looks like the flush puts output in correct order when I call the test your way. I will push that change and then I can merge?

golaz:
Thanks, go ahead and merge.

zshaheen:
Hi @golaz, when you say the below, isn't that what the tests *kind of* already do?

> I don't know if @zshaheen had the intention to add one more level of verification by separately computing md5sum and checking them against the database.

Specifically, [this line from this PR](https://github.com/E3SM-Project/zstash/pull/40/files#diff-b131fe2646b3e4980d06f90692dc2401R498). In that entire code block, we compute the MD5 hash for the local files and compare them against the MD5 in the database.

One note is that I was only printing the hashes, not checking/asserting the diff. If I remember correctly, it was cause of some issues with folders or symlinks having different hashes or something. You guys might wanna examine this.

If I'm misunderstanding, feel free to correct my understanding.

golaz:
Hi @zshaheen : thanks, that's what I thought and that makes sense. So far, we are only printing the hashes, not checking them in this part of the test. Indeed, there are some mismatches for directories and links, which we would have to factor in if we wanted to check them as part of the test.

PR #42, opened 2020-02-04
Option for unavailable HPSS

forsyth2:
Add functionality to allow Zstash to function on systems without HPSS.

Completes step 1 of 2 in #21, in addition to being the second and last pull request to replace #39.

Checklist of issues mentioned in #39:
- [x] `zstash check hpss=none` should NOT delete tar files.
- [x] When using `hpss=none` make the tar files read-only
- [x] Add test to check  that `zstash check hpss=none` does NOT delete tar files.
- [x] Remove all `hsi` operations from the non-HPSS tests
- [x] On non-HPSS systems, only run the non-HPSS tests
- [x] Fix `No such file or directory` error from `find zstash_test_backup -regex .*\.txt.* -exec md5sum {}`
- [x] Fix `chgrp` test
- [ ] Increment version number (this will be done after merging this pull request)

### DIFF
--- a/tests/test.py
+++ b/tests/test.py
@@ -44,14 +44,12 @@ def cleanup(hpss_path):
     After the script has failed/run, remove all created files, even those on the HPSS repo.
     """
     print('Removing test files, both locally and at the HPSS repo')
-    if os.path.exists('zstash_test'):
-        shutil.rmtree('zstash_test')
-    if os.path.exists('zstash_test_backup'):
-        shutil.rmtree('zstash_test_backup')
-    if os.path.exists('zstash'):
-        shutil.rmtree('zstash')
-    cmd = 'hsi rm -R {}'.format(hpss_path)
-    run_cmd(cmd)
+    for d in ['zstash_test', 'zstash_test_backup', 'zstash']:
+        if os.path.exists(d):
+            shutil.rmtree(d)
+    if hpss_path and hpss_path.lower() != 'none':
+        cmd = 'hsi rm -R {}'.format(hpss_path)
+        run_cmd(cmd)
 
 
 # Compare content of two (unordered lists)
@@ -60,7 +58,7 @@ def compare(s, t):
     return Counter(s) == Counter(t)
 
 
-class TestZStash(unittest.TestCase):
+class TestZstash(unittest.TestCase):
     def stop(self, hpss_path, error_message):
         """
         Cleanup and stop running this script.
@@ -90,6 +88,7 @@ class TestZStash(unittest.TestCase):
             print('*' * 40)
             self.stop(hpss_path, error_message)
 
+    @unittest.skipIf(os.system('which hsi') != 0, 'This system does not have hsi')
     def testZstashWithHPSS(self):
         print('*' * 40)
         print('testZstashWithHPSS')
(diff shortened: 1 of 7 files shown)


forsyth2:
@golaz I still need to run the tests on Compy, but how do these changes look? The tests pass on Cori.

zstash/extract.py
```diff
@@ -139,7 +139,11 @@ def extract(keep_files=True):
     config.keep = bool(int(config.keep))
 
     # The command line arg should always have precedence
-    config.keep = args.keep
+    if args.hpss == 'none':
+        # If no HPSS is available, always keep the files.
+        config.keep = True
```
forsyth2: This being set to `False` is what causes the tar file deletion. I suppose we could run `os.chmod` on the tar file if `config.keep == True` to make the tar file read-only in this case, but this if-block appears to be sufficient.

tests/test.py
```diff
@@ -512,6 +514,124 @@ def testZstashWithHPSS(self):
 
         cleanup(hpss_path)
 
+    def testZstashWithoutHPSS(self):
```
forsyth2: Is this correctly testing everything we care about and nothing we don't care about?

tests/test.py
```diff
@@ -90,6 +91,7 @@ def str_in(self, cmd, hpss_path, output, msg):
             print('*' * 40)
             self.stop(hpss_path, error_message)
 
+    @unittest.skipIf(os.system('which hsi') != 0, 'This system does not have hsi')
```
forsyth2: Tested on Compy. This test is successfully skipped and the other two pass.

forsyth2:
@golaz Ready for review.

zstash/extract.py
```diff
@@ -388,6 +392,8 @@ def extractFiles(files, keep_files, keep_tars, multiprocess_worker=None):
             # Delete this tar if the corresponding command-line arg was used.
             if not keep_tars:
                 os.remove(tfname)
+            else:
+                os.chmod(tfname, 0o444)
```
forsyth2: Would this be sufficient to make tar-files read only if we're keeping them? Or would it be better to set it based on `config.hpss == 'none'` instead of `keep_tars == True`?

golaz: This will change default behavior even when using HPSS. I would prefer that it only applied when not using HPSS.

zstash/extract.py
golaz: Maybe add 'lower' to the if test just in case someone specifies hpss=None.

golaz: Actually, since you are testing against 'none' other places in the code, we should make sure to lower capitalization of the command line argument.

tests/test.py
golaz: - Can you add a test for 'zstash extract' in testZstashWithoutHPSS?
- In 'testKeepTars' I would add the following steps
  - delete files (but not zstash files)
  - extract everything, make sure there are not errors
  - make sure tar files are still there
  - check that tar files are read-only

zstash/extract.py
forsyth2: @golaz Can that be done in one file (i.e., `main.py`) or would I have to apply `.lower()` in all the python files representing commands?

I was thinking it would be good to simply have `args.hpss` default to the `None` type if the user doesn't specify `--hpss`. However, `chgrp` appears to be an obstacle to that idea because `hpss` is passed in by position rather than command line option; changing how `chgrp` is called would break backward compatibility.

tests/test.py
forsyth2: @golaz What is the expected behavior of `zstash extract` if there is no HPSS available to extract from? When I try running `zstash extract` I get `sqlite3.OperationalError: unable to open database file`.

zstash/extract.py
golaz: @forsyth2 : yes, it probably would be cleaner to do the lower() in main.py rather than everywhere it appears in an if statement. config.hpss could be set to `None` rather than a string "none".

Setting args.hpss to default to `None` is likely break other things. In particular, you can invoke zstash without `--hpss` and in that case zstash should use the path stored in the database. The idea was to avoid having to specify `--hpss` at every invocation:

```
zstash create --hpss=... ...
zstash update   # no need to repeat --hpss
```

The 'chgrp' option is rarely used (and apparently not mentioned in the documentation pages), so you can break backwards compatibility for it.

tests/test.py
golaz: zstash extract (with the zstash/ subdirectory present and containing index.db and tar files) should extract all the files from the local tar files.

tests/test.py
```diff
@@ -512,6 +514,457 @@ def testZstashWithHPSS(self):
...
+        output, err = run_cmd(cmd)
+        os.chdir('../')
+        self.str_in(cmd, hpss_path, output+err, 'Not extracting file0.txt')
+        self.str_in(cmd, hpss_path, output+err, 'Not extracting file0_hard.txt')
+        # It's okay to extract the symlinks.
+        self.str_not_in(cmd, hpss_path, output+err, 'Not extracting file0_soft.txt')
+        self.str_in(cmd, hpss_path, output+err, 'Not extracting file_empty.txt')
+        self.str_in(cmd, hpss_path, output+err, 'Not extracting dir/file1.txt')
+        # It's okay to extract empty dirs.
+        #self.str_not_in(cmd, hpss_path, output+err, 'Not extracting empty_dir')
```
forsyth2: Interestingly, this fails on Compy, but not Cori. If we're ok with having "not extracting empty_dir" in `output+err`, then I can remove this commented assertion.

tests/test.py
```diff
@@ -512,6 +514,457 @@ def testZstashWithHPSS(self):
...
+        output, err = run_cmd('zstash extract --hpss={}'.format(hpss_path))
+        # Run `zstash check`
+        output, err = run_cmd('zstash check --hpss={}'.format(hpss_path))
+        self.assertEqual(output+err, 'INFO: Opening tar archive zstash/000000.tar\nINFO: Checking file1.txt\nINFO: Checking file2.txt\nINFO: No failures detected when checking the files.\n')
+        # Check that tar and db files were not deleted
+        actual = sorted(os.listdir('zstash/'))
+        expected = sorted(['000000.tar', 'index.db'])
+        self.assertEqual(actual, expected)
+        # Check that tar file is read-only
+        # https://stackoverflow.com/questions/1861836/checking-file-permissions-in-linux-with-python
```
forsyth2: Ok to keep link?

tests/test.py
```diff
@@ -512,6 +514,457 @@ def testZstashWithHPSS(self):
...
+        stat = os.stat('zstash/000000.tar')
+        oct_mode = str(oct(stat.st_mode))[-3:]
+        self.assertEqual(oct_mode, '444')
```
forsyth2: I think there are named flags in https://docs.python.org/3/library/os.html#os.chmod but I figured `444` was fine for read-only for everyone.

forsyth2:
@golaz Ready for code review again. All tests pass on both Cori and Compy.

golaz:
@forsyth2 : getting really close. I checked out and installed your branch on cori and ran the standard tests with no problems. I performed a few more tests myself and found two minor changes to make:

- The 'chmod' command should be moved from `extract.py` to `create.py`. We should make the tar files read-only as soon as we close them. zstash will never append to an existing tar file, so this can be done right away.
- Change chmod to the equivalent Python equivalent of `chmod -w` (not sure what it is). Setting permissions to 444 will make the tar file world readable, which may not be what the user intended. (Sorry for not catching this when we discussed it earlier).

forsyth2:
Tests pass on both Cori and Compy.

golaz:
@forsyth2 : thanks for working on this. This will be a very useful new feature to have. Are you willing to also update the documentation?

forsyth2:
@golaz Yeah, I can do that. I'll make a new pull request off the `gh-pages` branch.

forsyth2:
Documentation in #45

PR #48, opened 2020-02-25
Add a conda build config for py 3.6, 3.7 and 3.8

xylar:
This will make it easier to build all supported python versions in one go.

### DIFF
--- a/conda/conda_build_config.yaml
+++ b/conda/conda_build_config.yaml
@@ -1,0 +1,10 @@
+python:
+  - 3.6
+  - 3.7
+  - 3.8
+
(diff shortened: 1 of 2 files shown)


chengzhuzhang:
@xylar Thank you for the PR! We also would like to make zstash build noarch. Not sure how to edit the yml file, would you please help...Thanks a lot!

xylar:
Ah, to make it noarch, you don't want the this PR at all.  I'll close and do noarch instead.  Give me an hour or so.  Off to a meeting...

chengzhuzhang:
@xylar That would be great! Thank you.

PR #50, opened 2020-02-25
Switch zstash build to noarch

xylar:
Also a lot of other clean-up, see comments below.

### DIFF
--- a/conda/meta.yaml
+++ b/conda/meta.yaml
@@ -22,6 +22,12 @@ requirements:
   run:
     - python >=3.6
 
+test:
+  imports:
+    - zstash
+  commands:
+    - zstash --help
+
 about:
   home: https://github.com/E3SM-Project/zstash
 
--- a/conda/meta.yaml
+++ b/conda/meta.yaml
@@ -1,22 +1,27 @@
+{% set name = "zstash" %}
+{% set version = "0.4.0" %}
+
 package:
-    name: zstash
-    version: 0.4.0
+  name: {{ name|lower }}
+  version: {{ version }}
 
 source:
-    git_url: https://github.com/E3SM-Project/zstash
-    git_rev: v0.4.0
+  url: https://github.com/E3SM-Project/{{ name }}/archive/v{{ version }}.tar.gz
+  sha256: 44d0607e4826792329c37cc68efc647ce78a0263a9f811319a343bd91e9d1d8e
 
 build:
-    script: python setup.py install
+  number: 1
+  script: "{{ PYTHON }} -m pip install . --no-deps -vv"
+  noarch: python
 
 requirements:
-  build:
-    - python 3
-    - setuptools
+  host:
+    - python >=3.6
+    - pip
 
   run:
-    - python 3
+    - python >=3.6
 
 about:
-    home: https://github.com/E3SM-Project/zstash
+  home: https://github.com/E3SM-Project/zstash
 


conda/meta.yaml
```diff
@@ -1,22 +1,27 @@
+{% set name = "zstash" %}
+{% set version = "0.4.0" %}
```
xylar: These Jinja2 variables let you reuse the same name and version in multiple places.  Standard on conda-forge.

conda/meta.yaml
```diff
@@ -1,22 +1,27 @@
...
-    name: zstash
-    version: 0.4.0
+  name: {{ name|lower }}
+  version: {{ version }}
 
 source:
-    git_url: https://github.com/E3SM-Project/zstash
-    git_rev: v0.4.0
+  git_url: https://github.com/E3SM-Project/{{ name }}/archive/v{{ version }}.tar.gz
+  sha256: 9fce7561efa0d512567f89219a1f1642210810523a550b1974a17fdba1b6f544
```
xylar: Better, faster and safer to use the tar file than the git tag.  This way, if the file changes you find out that the hash is "wrong".

conda/meta.yaml
```diff
@@ -1,22 +1,27 @@
...
 
 build:
-    script: python setup.py install
+  number: 1
```
xylar: Very important to have a build number.  I set it to 1 because build 0 has already been uploaded.  Never overwrite a build unless you have a very good reason to do so.

conda/meta.yaml
```diff
@@ -1,22 +1,27 @@
...
+  script: "{{ PYTHON }} -m pip install . --no-deps -vv"
```
xylar: This is the standard install command on conda-forge, better than `python setup.py install`.  Ask or Google if you want details on why.

conda/meta.yaml
```diff
@@ -1,22 +1,27 @@
...
+  noarch: python
```
xylar: The actual point of the PR...

conda/meta.yaml
```diff
@@ -1,22 +1,27 @@
...
 
 requirements:
-  build:
-    - python 3
-    - setuptools
+  host:
+    - python >=3.6
+    - pip
```
xylar: the new install command uses pip, rather than setuptools.

conda/meta.yaml
```diff
@@ -1,22 +1,27 @@
...
+  number: 1
+  script: "{{ PYTHON }} -m pip install . --no-deps -vv"
+  noarch: python
 
 requirements:
-  build:
-    - python 3
-    - setuptools
+  host:
+    - python >=3.6
```
xylar: Here and below, it is more typical to constrain python in this way, though you could also specify that it has to be less than 4 if you have reason to be worried.

xylar:
A couple of bugs.  Stand by...

chengzhuzhang:
@xylar Thank you for the PR and clarification on the changes! 👍 
@forsyth2 Can you please try upload a new build with the updated recipe?

forsyth2:
@chengzhuzhang I installed the latest Zstash using Conda on both Compy and my Mac. `zstash -h` shows `v0.4.0`.

chengzhuzhang:
@forsyth2 Hi, it doesn't seem a new build is uploaded to anaconda... https://anaconda.org/e3sm/repo/history  I think you want to merge this PR and follow through the `release to anaconda` process again.

forsyth2:
@xylar @chengzhuzhang I ran the tests on Cori using the new environment with latest Python and Python 3.6. Tests pass, so I think this is good to merge. Then I'll upload to e3sm conda channel.

xylar:
Great, go ahead and merge when you want.

forsyth2:
@xylar @chengzhuzhang @golaz Ok package updated on https://anaconda.org/e3sm/zstash

PR #55, opened 2020-02-27
Fix tar deletion bug

forsyth2:
Fix tar deletion bug - this pull request resolves #53 . Also 1) add ZSTASH_DIR variable to tests to allow user to specify zstash installation location and 2) change group to e3sm.

After this merges, I'll follow the release steps to release v0.4.1 on GitHub and Anaconda

PR #56, opened 2020-02-27
Update to v0.4.1

forsyth2:
Update to v0.4.1 following bug fix in #55.


conda/meta.yaml
```diff
@@ -1,5 +1,5 @@
 {% set name = "zstash" %}
-{% set version = "0.4.0" %}
+{% set version = "0.4.1" %}
```
forsyth2: @xylar Does the `sha256` on line 10 also need to change if we increment the version?

forsyth2: Also, should the build number on line 13 be incremented on every "update version" commit?

xylar: Each time you change the version, you need to reset the build number to zero and change the sha256 hash.  (The easiest way to do the latter is just to run `conda build .` and wait until it complains that the hashes don't match and use that error message to find the new hash.  But the "right" way is to download the tar file and run `sha256sum` on it.

xylar: Also, on `conda-forge`, the feedstock is separate from the repo.  In this case where the recipe is in the repo, you have the awkward situation where you do a release and then you have to update the recipe only after the release because you have no other way to get the `sha256` hash.

Issue #81, opened 2020-08-10
Update installation directions


forsyth2:
`source /global/project/projectdirs/acme/software/anaconda_envs/load_latest_e3sm_unified.sh` on the "Getting Started" page should be changed to `source /global/cfs/cdirs/e3sm/software/anaconda_envs/load_latest_e3sm_unified.sh` for Cori and `source /share/apps/E3SM/conda_envs/load_latest_e3sm_unified.sh` for Compy.

forsyth2:
Merging #92 should have closed this.

Issue #78, opened 2020-07-28
Update docs with options


forsyth2:
In the docs, not all available options for each `zstash` command are described. `--keep` needs to be added for `check` and `update`. `-v` needs to be added for `create`, `check`, `update`, `extract`, and `ls`.

forsyth2:
#94 merged, so we can close this issue.

PR #85, opened 2020-08-18
Add versions to documentation

### LINKED ISSUE #71: Documentation versions
Find a way to have multiple versions of documentation online. Suppose version `n` is the latest release but `master` has more commits and the docs have been updated accordingly. Currently, users would see the latest docs online even though the docs describe features not included in the version they are using.

Possible solution would be to just have the `index.html` point to a bulleted list of docs (v1, v2, master/merged-but-not-released-yet).

We can also look into updating the documentation in `master` instead of on `gh-pages`. This would be nice because developers could update code and documentation in the same PR. However, it wouldn't solve the primary problem of needing to serve multiple versions of the docs online (unless we had users refer to the docs included in the release they download rather than looking online).

forsyth2:
Resolves #71. Add versions to documentation. This is accomplished by having https://e3sm-project.github.io/zstash/docs/html/index.html be the latest documentation (e.g., documentation reflecting `master`).

The documentation currently reflects v0.4.2, which I just released today. I believe the easiest way to version the documentation is by doing the following:

```
conda activate sphinx
make html
cp docs/html docs/html-v0-4-2
```

`docs/html/` is the latest documentation that will continue to be updated. `docs/html-v0-4-2` is the documentation from v0.4.2 that will now be left alone. So, the `source` directory will *only* include the latest documentation. The `docs` directory will hold "snapshots" of the `html` from each release.

On `source/index.rst` I have added a bulleted list of previous documentation versions. Right now, it only contains the v0.4.2 docs. Going forward we can add to that list with each release. I used explicit links rather than `:ref` references because the `html` directories are essentially independent and I don't think the documentation generator is able to connect them. (As a result, the link is broken for me because https://e3sm-project.github.io/zstash/docs/html-v0-4-2/index.html doesn't exist yet).

If this versioning method is acceptable, I can update the "How to Prepare a Release" docs to add a new documentation version.


forsyth2:
@golaz What do you think of the versioning method I described above? If you think it's acceptable I will update the "How to Prepare a Release" docs to reflect this process and merge this pull request.

@chengzhuzhang I think this versioning method would also work out well for the E3SM Diags docs.

forsyth2:
It occurs to me we could probably also make copies of `source`, so that we could update documentation for previous releases (e.g., if the documentation was discovered to be incorrect).

forsyth2:
The second commit includes a copy of `source` and an update to the docs on releasing a new version. I will combine the two commits before merging.

forsyth2:
It would probably be good to make sure the v0.4.2 docs are good before copying over the documentation as a previous version. (i.e. would be good to merge #78, #81 before this one).

chengzhuzhang:
@forsyth2 Thanks for working on this! I like the versioning method you proposed here. And this implies to make `update documentation` a standard practice each time before a code release with new features.

forsyth2:
@chengzhuzhang great! When I have time, I will implement something similar for E3SM Diags to address https://github.com/E3SM-Project/e3sm_diags/issues/299.

forsyth2:
Since more docs updates have been added that are for v0.4.2 since this was created, it's cleaner to close this pull request in favor of #95.

Issue #88, opened 2020-08-21
Conda package of zstash 0.4.2 incorrectly includes python_abi


xylar:
```
zstash 0.4.2 py_0
-----------------
file name   : zstash-0.4.2-py_0.tar.bz2
name        : zstash
version     : 0.4.2
build       : py_0
build number: 0
size        : 20 KB
subdir      : noarch
url         : https://conda.anaconda.org/e3sm/noarch/zstash-0.4.2-py_0.tar.bz2
md5         : b284b7db42dd55f4eb60cef99d84f4f7
timestamp   : 2020-08-17 21:05:36 UTC
dependencies: 
  - python >=3.6
  - python_abi 3.8.* *_cp38
```
As a result, `zstash` cannot be installed for any version of python other than 3.8.  E3SM-Unified supports only python 3.6 and 3.7 at the moment.

@chengzhuzhang has also had this problem building packages of `e3sm_diags`.  I do not know what is incorrectly configured in your `conda build` that means `python_abi` is included in the package dependencies.  This does not happen for me.  I will see if I can reproduce this problem building on a Mac in case that is the issue.

xylar:
This issue https://github.com/conda/conda-build/issues/3956 suggests that the problem might be that you are using `conda-build` form defaults whereas you really need the version from conda-forge.  I would suggest setting channels to strict and doing a conda update on `conda` and `conda-build` to make sure you have the right version:
```
conda config --add channels conda-forge
conda config --set channel_priority strict
conda update -n base conda conda-build
```
@chengzhuzhang, can you try this as well and see if it fixes the build problem with `e3sm_diags`?

Once you have built the package locally, use:
```
conda search --info --use-local zstash
```
to make sure `python_abi` is not being included.

xylar:
I built `zstash` on a Mac and it still doesn't have `python_abi` for me, so I don't think it's a platform issue.  I think it's an issue of `conda-build` coming from the wrong channel as stated above.  Please always use strict channel priority and make sure `conda-forge` is the highest priority channel.

xylar:
I am able to build `zstash` without `python_abi` on Linux as well.  Here is my base environment:
```
$ conda list
# packages in environment at /home/xylar/miniconda3:
#
# Name                    Version                   Build  Channel
_libgcc_mutex             0.1                 conda_forge    conda-forge
_openmp_mutex             4.5                       1_gnu    conda-forge
alabaster                 0.7.12                     py_0    conda-forge
[...]

PR #90, opened 2020-08-21
 Bump build number to 1

xylar:
I'm just bumping the build number so we can do a new build and upload without `python_abi`.


xylar:
@forsyth2, could you try building again following my suggestions in #88?  If you're still getting `python_abi` as a dependency, please don't upload.  Let's try to debug and in the worst case, I'll upload build 1 since I'm able to build without the `python_abi` issue.

PR #91, opened 2020-08-24
Update Release Documentation

forsyth2:
Update documentation to improve the release process, based on suggestions in #88.


source/release.rst
```diff
@@ -118,16 +118,9 @@ Since we're building with ``noarch``, you can run the below steps on
...
-use ``source deactivate``. This means you have an older version of Anaconda,
-which should be remedied after the following ``update`` command.
-
-    ::
-
-        conda deactivate
-        conda update anaconda conda conda-build
+1. Be in the ``base`` conda environment.
+Make sure you have the latest versions of ``conda``, and ``conda-build``
+by running ``conda update -n base conda conda-build``.
```
forsyth2, line 123: Not entirely sure this is the correct update to this step.

forsyth2:
@xylar I updated the docs to improve the release process, based on your suggestions in #88. I also included a note about properly updating the E3SM Unified page. If you'd like to see the changes, you can simply view `source/release.rst`. Thanks!

@golaz Just a FYI that these steps are being updated.

source/release.rst
xylar: Yes, this is correct.

xylar: You should add `anaconda-client` to the list.  This is the package that supports the `anaconda upload` command.

xylar:
@forsyth2, somewhere in "Preparing The Code For Release", you need to say that the build number in the conda recipe needs to get set back to 0.

source/release.rst
```diff
@@ -118,16 +118,9 @@ Since we're building with ``noarch``, you can run the below steps on
...
-You cannot be in an existing Anaconda environment when you run ``conda update``,
-so run ``conda deactivate`` first. If the ``conda deactivate`` command doesn't work,
-use ``source deactivate``. This means you have an older version of Anaconda,
-which should be remedied after the following ``update`` command.
-
-    ::
-
-        conda deactivate
-        conda update anaconda conda conda-build
+1. Be in the ``base`` conda environment.
```
xylar, line 121:
```suggestion
1. Activate the ``base`` conda environment. if you have not already done so:

    ::
    
            conda activate base
```

source/release.rst
```diff
@@ -143,17 +136,26 @@ Or:
         git fetch origin master
         git checkout -b <branch-name> origin/master
 
-3. Run the command below. The ``conda/`` folder is where ``meta.yaml`` is located and the ``-c``
-tells conda the channels where the dependencies defined in ``meta.yaml`` can be found.
+3. Run ``conda env list``. There will be several paths listed.
+For each of those paths, run ``rm -rf <path>/miniconda3/conda-bld`` to ensure that you're building the latest package:
```
xylar, line 140: This is not necessary.  You only need to delete the `conda-bld` environment associated with your `miniconda3` installation.  There isn't one per environment.  The only reason you might see this in your particular case is because you have an installation on `anaconda3` and of `miniconda3`, which are completely separate from one another.

```suggestion
3. Run ``conda env list``. Determine the path for the `miniconda3` installation you are using to build the package. 
Typically, this will be `~/miniconda3`.  Run ``rm -rf <path>/conda-bld`` to ensure that previously built packages are not included in the current build.
```

source/release.rst
```diff
@@ -143,17 +136,26 @@ Or:
...
+
+4. Run the following commands:
```
xylar, line 142:
```suggestion
4. Run the following commands to make sure the `conda-forge` channel is included by default and that packages
come from that channel whenever possible:
```

source/release.rst
```diff
@@ -143,17 +136,26 @@ Or:
...
+5. Run ``conda build conda/ -c conda-forge``. The ``conda/`` folder is where ``meta.yaml`` is located and the ``-c``
+tells conda the channels where the dependencies defined in ``meta.yaml`` can be found. Keep the output of this command.
+We'll use it in step 7.
+
+6. Run ``conda search --info --use-local zstash``. The only dependency should be ``python >=3.6``. In particular,
+``python_abi`` should not be listed as a dependency.
 
+7. In the output of step 5, you should see something like the below.
+We only have one package of type ``noarch``, so it's compatible with both Python 2 and 3.
+Since we only officially support Python 3, it might not work with Python 2.
```
xylar, line 158:
```suggestion
7. In the output of step 5, you should see something like the below.
We only have one package of type ``noarch``, meaning it works on both Linux and OSX and is compatible with multiple versions of Python (3.6, 3.7, 3.8, etc.). 
Since we have constrained python versions to >= 3.6 in the dependencies, it will not work with Python 2 or any other version of Python <= 3.5.
```

xylar (approved):
I have some suggested changes above but I will leave the rest up to you.  The gist is certainly correct and should be a very useful update to the documentation. 

One other suggestion would be to say somewhere that users probably should start with an installation of Miniconda3 rather than Anaconda3 to build packages, including the link for where to download Miniconda3.  You can explain that all of the packages from the base Anaconda3 environment will come from the `defaults` channel, not the `conda-forge` channel.  Package updates fail when we try to update to the `conda-forge` version of `conda-build`,

Issue #96, opened 2020-09-29
Extract remaining files on second extract


forsyth2:
Currently, if `zstash extract` is used to only extract a subset of files, those files must be moved out of the directory before running again, or else a second `zstash extract` will not actually extract anything. The files not yet extracted should be extracted, however.

Example:
```
# On Compy
# Load unified environment
source /share/apps/E3SM/conda_envs/load_latest_e3sm_unified.sh
# Confirm we're on v0.4.2
zstash version
# Create directory
mkdir test_case
cd test_case
emacs aa0.txt
cp aa0.txt aa1.txt
cp aa0.txt aa2.txt
cp aa0.txt aa3.txt
cp aa0.txt aa4.txt
cp aa0.txt ab5.txt
cp aa0.txt ab6.txt
cp aa0.txt ab7.txt
cp aa0.txt ab8.txt
cp aa0.txt ab9.txt
# Create the zstash local cache
cd ..
zstash create --hpss=none test_case
# Create extraction directory
mkdir extraction
cd extraction
cp -r ../test_case/zstash/ zstash
# Extract a subset of files
zstash extract --hpss=none ab*
# Show that we've extracted all `ab` files
ls
# Extract remaining files
zstash extract --hpss=none a*
# ERROR: Remaining files NOT extracted!
ls
```

Add to test suite:
-`zstash extract --hpss=none ab*` -> `zstash extract --hpss=none a*`: remaining files should get extracted.

golaz:
@forsyth2, I figured out what's happening: the shell is playing tricks with us.

For the second extraction, replace

`zstash extract --hpss=none a*`

with

`zstash extract --hpss=none "a*"`

 and you'll get the expected behavior.

Ideally, we'd like to turn off shell substitution for zstash extract. I don't know if it is possible.

golaz:
Here are some possibilities, but they may not be very practical:

https://stackoverflow.com/questions/11456403/stop-shell-wildcard-character-expansion

golaz:
This is another great example of shell wildcard substitution playing tricks on unsuspecting users. 

Unfortunately, there is nothing we can do within zstash, because the shell substitution takes place **before** zstash is actually called.

The best option is to educate users. The documentation has been updated and now mentions this in a number of places:

https://e3sm-project.github.io/zstash/docs/html/usage.html?highlight=wildcards

PR #97, opened 2020-09-30
Include warnings for workers and wildcards

forsyth2:
Add easily visible warnings to reinforce bolded notes on numbers of workers and the use of wildcards. The wildcard warning will prevent users from encountering the error in #96.


golaz (approved):
OK from looking at the rst file.

PR #98, opened 2020-10-01
Test wildcards

forsyth2:
Add test for wildcards.

### DIFF
--- a/tests/test_extract.py
+++ b/tests/test_extract.py
@@ -10,12 +10,12 @@ class TestExtract(TestZstash):
     """
     # `zstash extract` is tested in TestExtract and TestExtractParallel.
     # x = on, no mark = off, b = both on and off tested
-    # option | ExtractVerbose | ExtractKeep | ExtractCache | ExtractParallel |
-    # --hpss    |x|x|x|x|
-    # --workers | | | |x|
-    # --cache   | | |x| |
-    # --keep    | |x| | |
-    # -v        |x| | |b|
+    # option | ExtractVerbose | ExtractKeep | ExtractCache | ExtractWildcard | ExtractParallel |
+    # --hpss    |x|x|x|x|x|
+    # --workers | | | | |x|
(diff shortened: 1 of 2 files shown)


tests/test_extract.py
```diff
@@ -148,6 +187,9 @@ def testExtractKeepHPSS(self):
     def testExtractCache(self):
         self.helperExtractCache('testExtractCache', 'none')
 
+    def testExtractWildcard(self):
```
forsyth2, line 190: Do we need a test with HPSS too?

tests/test_extract.py
```diff
@@ -131,6 +131,45 @@ def helperExtractCache(self, test_name, hpss_path, zstash_path=ZSTASH_PATH):
...
+        os.mkdir(self.test_dir)
+        os.chdir(self.test_dir)
+        if not use_hpss:
+            shutil.copytree('{}/{}/{}'.format(TOP_LEVEL, self.backup_dir, self.cache), self.copy_dir)
+        cmd = '{}zstash extract --hpss={} ab*'.format(zstash_path, self.hpss_path)
+        output, err = run_cmd(cmd)
+        print(output + err)
+        expected_present = ['Extracting ab2.txt', 'Extracting ab3.txt']
+        self.check_strings(cmd, output + err, expected_present, [])
+        cmd = '{}zstash extract --hpss={} a*'.format(zstash_path, self.hpss_path)
```
forsyth2, line 164: Interestingly, adding double quotes seems to cause a failure and not adding them allows the tests to pass -- the reverse of our experience as users. I believe this is because `subprocess.Popen` requires `cmd` to be converted to a list and I imagine the conversion encapsulates `a*` in quotes. Not sure if there's a way to automate a test in such a way as to use the quotes.

forsyth2: Try using `shell=True`? If that doesn't work, perhaps we should just close #96 without adding automated testing.

PR #102, opened 2020-10-16
Add license

### LINKED ISSUE #89: zstash needs a LICENSE!
I'm pretty sure E3SM would want there to be some sort of LICENSE file.

forsyth2:
Add license for `zstash`. Resolves #89.

### DIFF
--- a/LICENSE
+++ b/LICENSE
@@ -1,0 +1,29 @@
+BSD 3-Clause License
+
+Copyright (c) 2018, E3SM:  Energy Exascale Earth System Model
+All rights reserved.
+
+Redistribution and use in source and binary forms, with or without
+modification, are permitted provided that the following conditions are met:
+
+* Redistributions of source code must retain the above copyright notice, this
+  list of conditions and the following disclaimer.
+
+* Redistributions in binary form must reproduce the above copyright notice,
+  this list of conditions and the following disclaimer in the documentation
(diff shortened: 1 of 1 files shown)


LICENSE
```diff
@@ -0,0 +1,29 @@
+BSD 3-Clause License
+
+Copyright (c) 2018, E3SM:  Energy Exascale Earth System Model
```
forsyth2, line 3: Is the copyright year for E3SM or for `zstash`? If the latter, should it be the year `zstash` was created or the year we're adding the license?

forsyth2 (commented):
Copied the license from E3SM Diags (https://github.com/E3SM-Project/e3sm_diags/blob/master/LICENSE). E3SM has a similar license (https://github.com/E3SM-Project/E3SM/blob/maint-1.0/LICENSE).

LICENSE
xylar: It should be for z-stash.  It should always be the current year or a range from the year that the code was started to the current year.

xylar: I should also say that you probably can't just add a license.  You may need to consult with LLNL about the details here.

golaz:
@forsyth2 : Dave is working on this with the lab. We should wait until that's complete.

forsyth2:
Closing in favor of #153.

Issue #67, opened 2020-04-03
Improve stopping of Zstash


forsyth2:
Ctrl-c does not always work to kill the Zstash process. Sometimes, users have to find the process id and use the `kill` command. It would be ideal if ctrl-c would behave as expected.

golaz:
The issue manifests itself for example when extracting files without using the parallel option and with tar files already in the cache (in parallel mode, CTRL-C seems to work fine). For example:

`zstash extract -v --keep --hpss=/home/t/tang30/2018/E3SM_simulations/repaired/20180129.DECKv1b_piControl.ne30_oEC.edison "*cam.h0.010*"`

CTRL-C appears to have no impact. The only option is to find the process number and then kill it.

Trying to implement a few options.

1. Based on https://stackoverflow.com/questions/21120947/catching-keyboardinterrupt-in-python-during-program-shutdown

```
import sys
import time

def main():

    while True:
        pass

if __name__ == '__main__':
    try:
        main()
    except KeyboardInterrupt:
        print('Interrupted')
        sys.exit(1)
```

2. Based https://www.devdungeon.com/content/python-catch-sigint-ctrl-c

```
from signal import signal, SIGINT
from sys import exit

def handler(signal_received, frame):
    # Handle any cleanup here
    print('SIGINT or CTRL-C detected. Exiting gracefully')
    exit(0)

if __name__ == '__main__':
    # Tell Python to run the handler() function when SIGINT is recieved
    signal(SIGINT, handler)

    print('Running. Press CTRL-C to exit.')
    while True:
        # Do nothing and hog CPU forever until SIGINT received.
        pass

```

while both options do something in stand-alone tests above, it's not obvious if the work within zstash. Cori scracth is no longer working properly, so further test will have to wait.

golaz:
Closed via https://github.com/E3SM-Project/zstash/pull/114

PR #114, opened 2021-02-18
CTRL-C to kill zstash

golaz:
CTRL-C does not always kill zstash; this PR should fix this problem.

Based on https://www.devdungeon.com/content/python-catch-sigint-ctrl-c but replacing

`sys.exit(1)`

with

`os._exit(1)`

as `sys.exit(1)` does not appear to be sufficient.

Addresses https://github.com/E3SM-Project/zstash/issues/67

### DIFF
--- a/zstash/main.py
+++ b/zstash/main.py
@@ -2,8 +2,10 @@
 from __future__ import print_function, absolute_import
 
 import argparse
+import os
 import os.path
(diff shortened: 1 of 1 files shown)


forsyth2 (approved):
I tested this on Cori and it appears to work.

Issue #104, opened 2020-10-19
Move documentation to master


forsyth2:
GitHub now allows documentation to be built from a branch other than `gh-pages`. If we build the docs off `master`, then developers will only need to submit one pull request for 1) new code, 2) unit tests, and 3) documentation, rather than submitting a separate documentation pull request to `gh-pages`. See https://docs.github.com/en/free-pro-team@latest/github/working-with-github-pages/configuring-a-publishing-source-for-your-github-pages-site

golaz:
There is definitely an advantage to having the documentation source files on master, rather that `gh-pages`. However, I'm concerned about having the html build files on master as well. I think this will greatly "pollute" any PR that also includes documentation changes.

I'm wondering if we could have the best of both worlds with a hybrid approach: documentation source code on master, documentation build html on `gh-pages`. The only downside would be a more complicated build process; but maybe that could be scripted or automated.

Also, I have been looking into how to maintain documentation for multiple version with sphinx. It seems that it might require using an extension. Some options:

- https://sphinxcontrib-versioning.readthedocs.io/en/latest/index.html
- https://holzhaus.github.io/sphinx-multiversion/master/index.html

Read-the-docs can also do documentation for multiple versions. But I don't like the idea of writing documentation so that a commercial service can place ads on them (even if the "ads [are] served ethically", whatever that means).

forsyth2:
From @tomvothecoder: if we stick with GitHub Pages, we will use GitHub Actions to build the docs on `gh-pages` after all of the tests pass. It will only push the built docs to the `gh-pages` branch.

Issue #103, opened 2020-10-19
Add documentation version dropdown


forsyth2:
At some point, to improve upon #71, it would be nice to have a dropdown box for versions. It would also be good to automatically include a banner specifying if the version is not the latest. Example: https://docs.bokeh.org/en/latest/

forsyth2:
https://docs.readthedocs.io/en/stable/versions.html

golaz:
From https://github.com/E3SM-Project/zstash/issues/104; I should have posted the second part here:

Also, I have been looking into how to maintain documentation for multiple version with sphinx. It seems that it might require using an extension. Some options:

- https://sphinxcontrib-versioning.readthedocs.io/en/latest/index.html
- https://holzhaus.github.io/sphinx-multiversion/master/index.html

Read-the-docs can also do documentation for multiple versions. But I don't like the idea of writing documentation so that a commercial service can place ads on them (even if the "ads [are] served ethically", whatever that means).

PR #120, opened 2021-03-01
Link latest documentation version v1.0.1

forsyth2:
Link latest documentation version (v1.0.1)


forsyth2:
@tomvothecoder Documentation for the latest version of `zstash` has been published. So, it's probably a good time to work on #104 (and #103). Although, I suppose it wouldn't hurt too much if we continue documentation content updates before you get to those issues.

PR #123, opened 2021-03-17
Remove MacOS from GH Actions build matrix

tomvothecoder:
This PR removes MacOS from the GH Actions build matrix since `zstash` doesn't run on MacOS.

### DIFF
--- a/.github/workflows/workflow.yml
+++ b/.github/workflows/workflow.yml
@@ -9,12 +9,7 @@ on:
 
 jobs:
   build:
-    strategy:
-      fail-fast: false
-      matrix:
-        os: ["ubuntu-latest", "macos-latest"]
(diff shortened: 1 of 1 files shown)


tomvothecoder:
@forsyth2 FYI, I had the GH Actions `publish-docs` step also run on PRs to `master`.  
By creating a PR to your **fork's** `master`, you can generate a preview of your branch's changes on your fork's GitHub Pages (e.g. https://tomvothecoder.github.io/zstash/).

However, if you create a PR to the **upstream** `master`, there is an authorization issue with the step where it tries to push to `gh-pages` so the build fails. I tried to find a conditional to run `publish-docs` only on PRs to your fork's `master`, but haven't found a way.

For now, we won't be able to generate live doc previews and can only do `.rst` reviews until a workaround is found.

Issue #134, opened 2021-04-01
Refactor `con` and `cur` variables from global to function arg


tomvothecoder:
The `con` and `cur` variables are only used in one function for each file that it exists in (`create.py`, `extract.py`, `ls.py`, and `update.py`), but is declared as a `global`.

```
   global con, cur
    con = sqlite3.connect(get_db_filename(cache), detect_types=sqlite3.PARSE_DECLTYPES)
    cur = con.cursor()
```

I also noticed some repeated logic involving database work. It might be better to declare the `con` and `cur` variables in `main.py` and pass them into each of the functions as an argument, rather than instantiating new `con` and `cur` variables in each function. I haven't looked too closely at those files to analyze the logic though, just a quick glance.

https://github.com/E3SM-Project/zstash/blob/master/zstash/main.py#L51
```
    if args.command == 'version':
        print(__version__)
    elif args.command == 'create':
        create()
    elif args.command == 'update':
        update()
    elif args.command == 'extract':
        extract()
    elif args.command == 'chgrp':
        chgrp()
    elif args.command == 'check':
        check()
    elif args.command == 'ls':
        ls()
    else:
        print('Unrecognized command')
        parser.print_help()
        sys.exit(1)
```

forsyth2:
I imagine at least some of this will be addressed as part of #138

PR #136, opened 2021-04-02
Address FIXME comments

### LINKED ISSUE #133: Address flake8 and mypy FIXME comments
This issue is related to PR #130

forsyth2:
Address FIXME comments. Resolves #133.

### DIFF
--- a/tests/base.py
+++ b/tests/base.py
@@ -177,8 +177,11 @@ class TestZstash(unittest.TestCase):
         """
         Set up directories for testing.
         """
-        # FIXME: Item "None" of "Optional[Any]" has no attribute "lower"mypy(error)
-        if self.hpss_path.lower() == "none":  # type: ignore
+        if self.hpss_path:
+            hpss_path = self.hpss_path
+        else:
+            raise Exception("Invalid self.hpss_path={}".format(self.hpss_path))
+        if hpss_path.lower() == "none":
             use_hpss = False
         else:
             use_hpss = True
--- a/zstash/create.py
+++ b/zstash/create.py
@@ -13,11 +13,9 @@ from .hpss_utils import add_files
 from .settings import DEFAULT_CACHE, config, get_db_filename, logger
 from .utils import exclude_files, run_command
 
-con = None
-cur = None
 
-
-def create():
+# FIXME: C901 'create' is too complex (19)
+def create():  # noqa: C901
 
     # Parser
     parser = argparse.ArgumentParser(
@@ -65,9 +63,7 @@ def create():
     # Copy configuration
     config.path = os.path.abspath(args.path)
     config.hpss = args.hpss
-    # FIXME: Incompatible types in assignment (expression has type "int", variable has type "None") mypy(error)
-    # Solution: https://stackoverflow.com/a/42279784
-    config.maxsize = int(1024 * 1024 * 1024 * args.maxsize)  # type: ignore
+    config.maxsize = int(1024 * 1024 * 1024 * args.maxsize)
     config.keep = args.keep
     if args.cache:
         cache = args.cache
@@ -78,13 +74,16 @@ def create():
     logger.debug("Running zstash create")
     logger.debug("Local path : %s" % (config.path))
     logger.debug("HPSS path  : %s" % (config.hpss))
-    logger.debug("Max size  : %i" % (config.maxsize))  # type: ignore
+    logger.debug("Max size  : %i" % (config.maxsize))
     logger.debug("Keep local tar files  : %s" % (config.keep))
 
     # Make sure input path exists and is a directory
     logger.debug("Making sure input path exists and is a directory")
-    # FIXME: Argument 1 to "isdir" has incompatible type "None"; expected "Union[str, bytes, _PathLike[str], _PathLike[bytes]]"mypy(error)
-    if not os.path.isdir(config.path):  # type: ignore
+    if config.path:
+        path = config.path
+    else:
+        raise Exception("Invalid config.path={}".format(config.path))
+    if not os.path.isdir(path):
         error_str = "Input path should be a directory: {}".format(config.path)
         logger.error(error_str)
         raise Exception(error_str)
@@ -105,8 +104,7 @@ def create():
 
     # Create cache directory
     logger.debug("Creating local cache directory")
-    # FIXME: Argument 1 to "chdir" has incompatible type "None"; expected "Union[int, Union[str, bytes, _PathLike[str], _PathLike[bytes]]]" mypy(error)
-    os.chdir(config.path)  # type: ignore
+    os.chdir(config.path)
     try:
         os.makedirs(cache)
     except OSError as exc:
@@ -123,7 +121,6 @@ def create():
     logger.debug("Creating index database")
     if os.path.exists(get_db_filename(cache)):
         os.remove(get_db_filename(cache))
-    global con, cur
     con = sqlite3.connect(get_db_filename(cache), detect_types=sqlite3.PARSE_DECLTYPES)
     cur = con.cursor()
 
@@ -163,23 +160,22 @@ create table files (
 
     # List of files
     logger.info("Gathering list of files to archive")
-    files: List[Tuple[str, str]] = []
+    file_tuples: List[Tuple[str, str]] = []
     for root, dirnames, filenames in os.walk("."):
         # Empty directory
         if not dirnames and not filenames:
-            files.append((root, ""))
+            file_tuples.append((root, ""))
         # Loop over files
         for filename in filenames:
-            files.append((root, filename))
+            file_tuples.append((root, filename))
 
     # Sort files by directories and filenames
-    files = sorted(files, key=lambda x: (x[0], x[1]))
+    file_tuples = sorted(file_tuples, key=lambda x: (x[0], x[1]))
 
     # Relative file path, eliminating top level zstash directory
-    # FIXME: Name 'files' already defined mypy(error)
-    files: List[str] = [  # type: ignore
+    files: List[str] = [
         os.path.normpath(os.path.join(x[0], x[1]))
-        for x in files
+        for x in file_tuples
         if x[0] != os.path.join(".", cache)
     ]
 
--- a/zstash/extract.py
+++ b/zstash/extract.py
@@ -12,7 +12,7 @@ import sys
 import tarfile
 import traceback
 from datetime import datetime
-from typing import List
+from typing import DefaultDict, List, Optional, Tuple
 
 from . import parallel
 from .hpss import hpss_get
@@ -39,15 +39,13 @@ def multiprocess_extract(num_workers, matches, keep_files, keep_tars, cache):
     # A dict of tar -> size of files in it.
     # This is because we're trying to balance the load between
     # the processes.
-    # FIXME: Need type annotation for 'tar_to_size' mypy(error)
-    tar_to_size = collections.defaultdict(float)  # type: ignore
+    tar_to_size_unsorted: DefaultDict[str, float] = collections.defaultdict(float)
     for db_row in matches:
         tar, size = db_row[5], db_row[2]
-        tar_to_size[tar] += size
+        tar_to_size_unsorted[tar] += size
     # Sort by the size.
-    # FIXME: Incompatible types in assignment (expression has type "OrderedDict[Any, Any]", variable has type "defaultdict[Any, Any]") mypy(error)
-    tar_to_size = collections.OrderedDict(  # type: ignore
-        sorted(tar_to_size.items(), key=lambda x: x[1])
+    tar_to_size: collections.OrderedDict[str, float] = collections.OrderedDict(
+        sorted(tar_to_size_unsorted.items(), key=lambda x: x[1])
     )
 
     # We don't want to instantiate more processes than we need to.
@@ -55,12 +53,11 @@ def multiprocess_extract(num_workers, matches, keep_files, keep_tars, cache):
 
     # For worker i, workers_to_tars[i] is a set of tars
     # that worker i will work on.
-    # FIXME: Need type annotation for 'workers_to_tars' mypy(error)
-    workers_to_tars = [set() for _ in range(num_workers)]  # type: ignore
+    workers_to_tars: List[set] = [set() for _ in range(num_workers)]
     # A min heap, of (work, worker_idx) tuples, work is the size of data
     # that worker_idx needs to work on.
     # We can efficiently get the worker with the least amount of work.
-    work_to_workers = [(0, i) for i in range(num_workers)]
+    work_to_workers: List[Tuple[int, int]] = [(0, i) for i in range(num_workers)]
     heapq.heapify(workers_to_tars)
 
     # Using a greedy approach, populate workers_to_tars.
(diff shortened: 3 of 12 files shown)


zstash/extract.py
```diff
@@ -69,12 +66,13 @@ def multiprocess_extract(num_workers, matches, keep_files, keep_tars, cache):
         workers_work, worker_idx = heapq.heappop(work_to_workers)
         workers_to_tars[worker_idx].add(tar)
         # Add this worker back to the heap, with the new amount of work.
-        heapq.heappush(work_to_workers, (workers_work + tar_to_size[tar], worker_idx))
+        worker_tuple: Tuple[float, int] = (workers_work + tar_to_size[tar], worker_idx)
+        # FIXME: error: Cannot infer type argument 1 of "heappush"
```
forsyth2, line 70: Both of the arguments have type annotations. I can't seem to fix this error.

zstash/hpss.py
```diff
@@ -19,16 +19,14 @@ def hpss_transfer(hpss, file_path, transfer_type, cache, keep=None):
             # https://unix.stackexchange.com/questions/46915/get-the-chmod-numerical-value-for-a-file
             display_mode = "stat --format '%a' {}".format(file_path).split()
             output = subprocess.check_output(display_mode).strip()
-            # FIXME: On Python 3 '{}'.format(b'abc') produces "b'abc'", not 'abc'; use '{!r}'.format(b'abc') if this is desired behavior mypy(error)
-            logger.info("{} original mode={}".format(file_path, output))  # type: ignore
+            logger.info("{!r} original mode={!r}".format(file_path, output))
```
forsyth2, line 22: `'{}'.format(b'abc')` and `'{!r}'.format(b'abc')` both return `"b'abc'"` for me. I suppose just adding the `!r` is fine?

tests/base.py
```diff
@@ -177,8 +177,11 @@ def setupDirs(self, test_name):
         """
         Set up directories for testing.
         """
-        # FIXME: Item "None" of "Optional[Any]" has no attribute "lower"mypy(error)
-        if self.hpss_path.lower() == "none":  # type: ignore
+        if self.hpss_path:
+            hpss_path = self.hpss_path
+        else:
+            raise Exception("Invalid self.hpss_path={}".format(self.hpss_path))
```
forsyth2, line 183: I think `mypy` is smart enough to determine that after this exception, `self.hpss_path` can't be `None`, but I think using a new variable not of type `Optional` in the alternative case improves readability.

zstash/parallel.py
```diff
@@ -102,6 +101,7 @@ class PrintQueue(collections.deque):
         """
 
         def __init__(self):
+            # TODO: Created the class independently of the factory method?
```
forsyth2, line 104: It looks like `namedtuple` creates a new type? Perhaps this code needs to be handled differently to properly add type annotations?

forsyth2 (commented):
@tomvothecoder This is ready for review.

`git grep -n FIXME` currently shows:
```
zstash/create.py:17:# FIXME: C901 'create' is too complex (19)
zstash/extract.py:70:        # FIXME: error: Cannot infer type argument 1 of "heappush"
zstash/extract.py:114:# FIXME: C901 'extract' is too complex (19)
zstash/extract.py:296:# FIXME: C901 'extractFiles' is too complex (29)
zstash/parallel.py:105:            # FIXME: NamedTuple type as an attribute is not supported mypy(error)
zstash/update.py:18:# FIXME: C901 'update' is too complex (26)
```

The cyclomatic complexity errors will be fixed in #137. The two remaining errors I have left comments for.

zstash/extract.py
```diff
@@ -182,18 +182,23 @@ def extract(keep_files=True):
             cur.execute(u"select value from config where arg=?", (attr,))
             value = cur.fetchone()[0]
             setattr(config, attr, value)
-    # FIXME: No overload variant of "int" matches argument type "None" mypy(error)
-    config.maxsize = int(config.maxsize)  # type: ignore
-    # FIXME: Incompatible types in assignment (expression has type "bool", variable has type "None")mypy(error)
-    config.keep = bool(int(config.keep))  # type: ignore
+    if config.maxsize:
```
tomvothecoder, line 185: I noticed that this chunk of code isn't in `extract.py` like in `create.py`.
```
    optional.add_argument(
        "--maxsize",
        type=float,
        help="maximum size of tar archives (in GB, default 256)",
        default=256,
    )
```
I'm assuming `config.maxsize` will always be `None` since it is never assigned through the command line argument.

If so, do we need to check for `config.maxsize` still? We might get away with deleting this check altogether.

zstash/extract.py
```diff
@@ -12,7 +12,7 @@
 import tarfile
 import traceback
 from datetime import datetime
-from typing import List
+from typing import DefaultDict, List, Optional, Tuple
 
 from . import parallel
 from .hpss import hpss_get
```
tomvothecoder, line 18: For some reason I can't comment on chunks of code that are collapsed in the diff viewer.

Anyways, need to remove lines 28, 29, and 174 like the other files (related to `con` and `cur`)

zstash/update.py
```diff
@@ -88,16 +83,21 @@ def update():  # noqa: C901
             cur.execute(u"select value from config where arg=?", (attr,))
             value = cur.fetchone()[0]
             setattr(config, attr, value)
-    # FIXME: Incompatible types in assignment (expression has type "bool", variable has type "None") mypy(error)
-    config.maxsize = int(config.maxsize)  # type: ignore
-    # FIXME: Incompatible types in assignment (expression has type "bool", variable has type "None") mypy(error)
-    config.keep = bool(int(config.keep))  # type: ignore
+    if config.maxsize:
```
tomvothecoder, line 86: Same comment about if we need to check for `config.maxsize` since there is not `optional.add_argument` for `maxsize`

zstash/hpss.py
tomvothecoder: I think `mypy` is complaining to be explicit by adding `!r` if `"b'abc'"` is the expected output. Looks good to me

zstash/extract.py
```diff
@@ -69,12 +66,13 @@ def multiprocess_extract(num_workers, matches, keep_files, keep_tars, cache):
         workers_work, worker_idx = heapq.heappop(work_to_workers)
         workers_to_tars[worker_idx].add(tar)
         # Add this worker back to the heap, with the new amount of work.
-        heapq.heappush(work_to_workers, (workers_work + tar_to_size[tar], worker_idx))
+        worker_tuple: Tuple[float, int] = (workers_work + tar_to_size[tar], worker_idx)
+        # FIXME: error: Cannot infer type argument 1 of "heappush"
+        heapq.heappush(work_to_workers, worker_tuple)  # type: ignore
```
tomvothecoder, line 71: `heapq.heappush` isn't using type annotation, so  it defaults the arguments to `Any`. 

```
def heappush(heap, item):
    """Push item onto heap, maintaining the heap invariant."""
    heap.append(item)
    _siftdown(heap, 0, len(heap)-1)
```
https://github.com/python/cpython/blob/master/Lib/heapq.py#L130

Since you infer the type for `worker_tuple` as not `Any` , mypy throws `Cannot infer type argument 1 of "heappush"mypy(error)` 

For some reason, it only affects the `item` arg and not `heap` even though both args are type `Any`.

tomvothecoder: I think we can just use `# type: ignore` here

zstash/extract.py
```diff
@@ -365,16 +369,23 @@ def extractFiles(  # noqa: C901
 
         try:
             # Seek file position
-            # FIXME: Item "None" of "Optional[IO[bytes]]" has no attribute "seek" mypy(error)
-            tar.fileobj.seek(file_tuple[6])  # type: ignore
+            if tar.fileobj:
+                fileobj = tar.fileobj
+            else:
+                raise Exception("Invalid tar.fileobj={}".format(tar.fileobj))
+            fileobj.seek(file_tuple[6])
```
tomvothecoder, line 376: Does this work instead of declaring a `fileobj` var?

```
if tar.fileobj:
    fileobj.seek(file_tuple[6])
else:
    raise Exception("Invalid tar.fileobj={}".format(tar.fileobj))
```

zstash/extract.py
```diff
@@ -182,18 +182,23 @@ def extract(keep_files=True):
...
             value = cur.fetchone()[0]
             setattr(config, attr, value)
-    # FIXME: No overload variant of "int" matches argument type "None" mypy(error)
-    config.maxsize = int(config.maxsize)  # type: ignore
-    # FIXME: Incompatible types in assignment (expression has type "bool", variable has type "None")mypy(error)
-    config.keep = bool(int(config.keep))  # type: ignore
+    if config.maxsize:
+        maxsize = config.maxsize
+    else:
+        raise Exception("Invalid config.maxsize={}".format(config.maxsize))
```
tomvothecoder, line 188: You should use `raise ValueError(...)` to be more explicit on what the exception is. 

https://docs.python.org/3/library/exceptions.html#ValueError
https://stackoverflow.com/a/43605619

zstash/extract.py
```diff
@@ -182,18 +182,23 @@ def extract(keep_files=True):
...
+    config.maxsize = int(maxsize)
+    if config.keep:
+        keep = config.keep
+    else:
+        raise Exception("Invalid config.keep={}".format(config.keep))
```
tomvothecoder, line 193: You should use raise ValueError(...) to be more explicit on what the exception is.

https://docs.python.org/3/library/exceptions.html#ValueError
https://stackoverflow.com/a/43605619

zstash/extract.py
```diff
@@ -365,16 +369,23 @@ def extractFiles(  # noqa: C901
...
             # Get next member
             tarinfo = tar.tarinfo.fromtarfile(tar)
 
             if tarinfo.isfile():
                 # fileobj to extract
+                extracted_file = tar.extractfile(tarinfo)
+                if extracted_file:
+                    fin = extracted_file
+                else:
+                    raise Exception("Invalid extracted_file={}".format(extracted_file))
```
tomvothecoder, line 387: You should use raise ValueError(...) to be more explicit on what the exception is.

https://docs.python.org/3/library/exceptions.html#ValueError
https://stackoverflow.com/a/43605619

zstash/extract.py
```diff
@@ -457,8 +466,10 @@ def extractFiles(  # noqa: C901
 
             # Delete this tar if the corresponding command-line arg was used.
             if not keep_tars:
-                # FIXME: Argument 1 to "remove" has incompatible type "Optional[Any]"; expected "Union[str, bytes, _PathLike[str], _PathLike[bytes]]"mypy(error)
-                os.remove(tfname)  # type: ignore
+                if tfname:
+                    os.remove(tfname)
+                else:
+                    raise Exception("Invalid tfname={}".format(tfname))
```
tomvothecoder, line 472: You should use raise ValueError(...) to be more explicit on what the exception is.

https://docs.python.org/3/library/exceptions.html#ValueError
https://stackoverflow.com/a/43605619

zstash/update.py
```diff
@@ -88,16 +83,21 @@ def update():  # noqa: C901
...
+        maxsize = config.maxsize
+    else:
+        raise Exception("Invalid config.maxsize={}".format(config.maxsize))
```
tomvothecoder, line 89: You should use raise ValueError(...) to be more explicit on what the exception is.

https://docs.python.org/3/library/exceptions.html#ValueError
https://stackoverflow.com/a/43605619

zstash/update.py
```diff
@@ -88,16 +83,21 @@ def update():  # noqa: C901
...
+    config.maxsize = int(maxsize)
+    if config.keep:
+        keep = config.keep
+    else:
+        raise Exception("Invalid config.keep={}".format(config.keep))
```
tomvothecoder, line 94: You should use raise ValueError(...) to be more explicit on what the exception is.

https://docs.python.org/3/library/exceptions.html#ValueError
https://stackoverflow.com/a/43605619

zstash/create.py
```diff
@@ -78,13 +74,16 @@ def create():
...
     logger.debug("Keep local tar files  : %s" % (config.keep))
 
     # Make sure input path exists and is a directory
     logger.debug("Making sure input path exists and is a directory")
-    # FIXME: Argument 1 to "isdir" has incompatible type "None"; expected "Union[str, bytes, _PathLike[str], _PathLike[bytes]]"mypy(error)
-    if not os.path.isdir(config.path):  # type: ignore
+    if config.path:
+        path = config.path
+    else:
+        raise Exception("Invalid config.path={}".format(config.path))
```
tomvothecoder, line 85: You should use raise ValueError(...) to be more explicit on what the exception is.

https://docs.python.org/3/library/exceptions.html#ValueError
https://stackoverflow.com/questions/43605512/python-difference-between-valueerror-and-exception

zstash/extract.py
```diff
@@ -203,8 +208,7 @@ def extract(keep_files=True):
     logger.debug("Running zstash " + cmd)
     logger.debug("Local path : %s" % (config.path))
     logger.debug("HPSS path  : %s" % (config.hpss))
-    # FIXME: Incompatible types in string interpolation (expression has type "None", placeholder has type "Union[int, float, SupportsInt]") mypy(error)
-    logger.debug("Max size  : %i" % (config.maxsize))  # type: ignore
+    logger.debug("Max size  : %i" % (config.maxsize))
```
tomvothecoder, line 211: If we remove the check for `config.maxsize`, we can also remove this line.

zstash/hpss_utils.py
```diff
@@ -45,8 +45,11 @@ def add_files(cur, con, itar, files, cache):
         # Close tar archive if current file is the last one or adding one more
         # would push us over the limit.
         next_file_size = tar.gettarinfo(current_file).size
-        # FIXME: Unsupported operand types for > ("int" and "None") mypy(error)
-        if i == nfiles - 1 or tarsize + next_file_size > config.maxsize:  # type: ignore
+        if config.maxsize:
+            maxsize = config.maxsize
+        else:
+            raise Exception("Invalid config.maxsize={}".format(config.maxsize))
```
tomvothecoder, line 51: You should use raise ValueError(...) to be more explicit on what the exception is.

zstash/parallel.py
tomvothecoder: You've probably seen this:

>This is expected (although the error message could be better). The NamedTuple constructor isn't meant to be used at runtime in a class constructor; instead you should define your NamedTuple type in the module scope.

https://github.com/python/mypy/issues/7531

zstash/ls.py
```diff
@@ -77,10 +76,16 @@ def ls():
...
             value = cur.fetchone()[0]
             setattr(config, attr, value)
-    # FIXME: No overload variant of "int" matches argument type "None" mypy(error)
-    config.maxsize = int(config.maxsize)  # type: ignore
-    # FIXME: No overload variant of "int" matches argument type "None" mypy(error)
-    config.keep = bool(int(config.keep))  # type: ignore
+    if config.maxsize:
+        maxsize = config.maxsize
+    else:
+        raise Exception("Invalid config.maxsize={}".format(config.maxsize))
```
tomvothecoder, line 82: You should use raise ValueError(...) to be more explicit on what the exception is.

zstash/ls.py
```diff
@@ -77,10 +76,16 @@ def ls():
...
+    config.maxsize = int(maxsize)
+    if config.keep:
+        keep = config.keep
+    else:
+        raise Exception("Invalid config.keep={}".format(config.keep))
```
tomvothecoder, line 87: You should use raise ValueError(...) to be more explicit on what the exception is.

tomvothecoder (changes requested):
Overall, looks good with some minor comments and questions. Happy to see type annotations!

tomvothecoder (commented):
Overall, looks good with some minor comments.

Also, I noticed you worked on `cur` and `con` stuff.  #134 would be good to address after this PR is merged.

zstash/extract.py
forsyth2: Isn't the whole point of `Any` that any type can be used though?

forsyth2:
Closing this pull request. Changes (including addressing the comments) will be included as part of #138.

forsyth2:
Closing this pull request. Changes (including addressing the comments) will be included as part of #138.

PR #140, opened 2021-04-19
Add checksums for tar files

### LINKED ISSUE #62: Add checksums for tar files
In a use case where a user needs to work on a remote machine on different sets of files (unknown upfront) extracted the same zstash tarballs, the user has to extract all files and transfer them to the remote machine. It will be more efficient to transfer the entire zstash tarballs once and extract needed files on the remote machine. Unfortunately, the database does not provide checksums of the zstash tarballs to make sure that the tarballs are transferred from HPSS to a local machine, and then from the local machine to the remote machine without any errors.
It will be helpful to add checksums of the zstash tarballs to the database.

forsyth2:
Add checksum for tarball. Resolves #62 

- [x] `create` will set up a `tars` table
- [x] `ls` has an option to display the `tars` table
- [x] `update` will create a `tars` table if one doesn't exist. (Tested by adding an option to `create` to not set up the `tars` table).
- [x] Tests for `ls --tars`

### DIFF
--- a/zstash/extract.py
+++ b/zstash/extract.py
@@ -337,7 +337,7 @@ def extractFiles(  # noqa: C901
     the contents of what's in its print queue.
     """
     failures: List[FilesRow] = []
-    tfname: Optional[str] = None
+    tfname: str
     newtar: bool = True
     nfiles: int = len(files)
     if multiprocess_worker:
--- a/zstash/hpss_utils.py
+++ b/zstash/hpss_utils.py
@@ -12,7 +12,45 @@ import _hashlib
 import _io
 
 from .hpss import hpss_put
-from .settings import BLOCK_SIZE, TupleFilesRowNoId, config, logger
+from .settings import BLOCK_SIZE, TupleFilesRowNoId, TupleTarsRowNoId, config, logger
+from .utils import create_tars_table, tars_table_exists
+
+
+# Minimum output file object
+class HashIO(object):
+    def __init__(self, name: str, mode: str, do_hash: bool):
+        self.f: _io.BufferedWriter = open(name, mode)
+        self.hash: Optional[_hashlib.HASH]
+        if do_hash:
+            self.hash = hashlib.md5()
+        else:
+            self.hash = None
+        self.closed: bool = False
+        self.position: int = 0
+
+    def tell(self) -> int:
+        return self.position
+
+    def write(self, s):
+        self.f.write(s)
+        if self.hash:
+            self.hash.update(s)
+        self.position += len(s)
+
+    def md5(self) -> Optional[str]:
+        md5: Optional[str]
+        if self.hash:
+            md5 = self.hash.hexdigest()
+        else:
+            md5 = None
+        return md5
+
+    def close(self):
+        if self.closed:
+            return
+
+        self.f.close()
+        self.closed = True
 
 
 def add_files(
@@ -21,22 +59,24 @@ def add_files(
     itar: int,
     files: List[str],
     cache: str,
+    skip_tars_md5: bool = False,
 ) -> List[str]:
 
     # Now, perform the actual archiving
     failures: List[str] = []
-    newtar: bool = True
+    create_new_tar: bool = True
     nfiles: int = len(files)
     archived: List[TupleFilesRowNoId]
     tarsize: int
     tname: str
     tfname: str
+    tarFileObject: HashIO
     tar: tarfile.TarFile
     for i in range(nfiles):
 
-        # New tar archive in the local cache
-        if newtar:
-            newtar = False
+        # New tar in the local cache
+        if create_new_tar:
+            create_new_tar = False
             archived = []
             tarsize = 0
             itar += 1
@@ -46,7 +86,15 @@ def add_files(
             tfname = "{}.tar".format(tname)
             logger.info("Creating new tar archive {}".format(tfname))
             # Open that tar file in the cache
-            tar = tarfile.open(os.path.join(cache, tfname), "w")
+            do_hash: bool
+            if not skip_tars_md5:
+                # If we're not skipping tars, we want to calculate the hash of the tars.
+                do_hash = True
+            else:
+                do_hash = False
+            tarFileObject = HashIO(os.path.join(cache, tfname), "wb", do_hash)
+            # FIXME: error: Argument "fileobj" to "open" has incompatible type "HashIO"; expected "Optional[IO[bytes]]"
+            tar = tarfile.open(mode="w", fileobj=tarFileObject)  # type: ignore
 
         # Add current file to tar archive
         current_file: str = files[i]
@@ -66,8 +114,9 @@ def add_files(
                 offset,
             )
             archived.append(t)
-            # Increase tarsize by the size of the current file
-            tarsize += size
+            # Increase tarsize by the size of the current file.
+            # Use `tell()` to also include the tar's metadata in the size.
+            tarsize = tarFileObject.tell()
         except Exception:
             # Catch all exceptions here.
             traceback.print_exc()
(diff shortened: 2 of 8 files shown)


zstash/hpss_utils.py
```diff
@@ -87,7 +117,18 @@ def add_files(
...
 
-            # Transfer tar archive to HPSS
+            tar_md5: str = tarFileObject.md5()
+            tarFileObject.close()
+            # TODO: This function is called by create.py and update.py,
+            # so, we need to test both `create` and `update`.
+            # TODO: test that tars table can be added to existing zstash database
+
+            tar_tuple: Tuple[str, int, str] = (tfname, tarsize, tar_md5)
+            print("tar_md5 on hpss_utils={}".format(tar_md5))
```
forsyth2, line 127: Print `tfname` too. Keep this line for the user too, using `logger`.

zstash/extract.py
```diff
@@ -393,6 +393,24 @@ def extractFiles(  # noqa: C901
         # True if we should actually extract the file from the tar
         extract_this_file: bool = keep_files and should_extract_file(files_row)
 
+        # TODO: implement this proof-of-concept
```
forsyth2, line 396: Maybe we can implement this as a new feature later on. We could modify `zstash ls` to list the md5 values too. Maybe add a `--tars` option to that command.

zstash/hpss_utils.py
```diff
@@ -87,7 +117,18 @@ def add_files(
             logger.debug("Closing tar archive {}".format(tfname))
             tar.close()
 
-            # Transfer tar archive to HPSS
+            tar_md5: str = tarFileObject.md5()
+            tarFileObject.close()
+            # TODO: This function is called by create.py and update.py,
+            # so, we need to test both `create` and `update`.
+            # TODO: test that tars table can be added to existing zstash database
```
forsyth2, line 124: `update.py` needs to be updated to create the `tars` table if it doesn't exist. Add a test case for this too.

forsyth2:
@golaz This is ready for review. I've listed the necessary/completed tasks in the checklist above. All tests pass on Cori.

I've listed what `zstash ls` output looks like in 4 scenarios below. Let me know if each seems acceptable.

1. `zstash ls` after `create` with no `tars` table
```
$ zstash ls --hpss=none
file0.txt
file0_hard.txt
file0_soft.txt
file0_soft_bad.txt
file_empty.txt
dir/file1.txt
empty_dir
```

2. `zstash ls --tars` after `create` with no `tars` table (notice `"tars table does not exist"`)
```
$ zstash ls --tars --hpss=none
file0.txt
file0_hard.txt
file0_soft.txt
file0_soft_bad.txt
file_empty.txt
dir/file1.txt
empty_dir

tars table does not exist
```

3. `zstash ls --tars` after `update` added the `tars` table
```
$ zstash ls --tars --hpss=none
file0.txt
file0_hard.txt
file0_soft.txt
file0_soft_bad.txt
file_empty.txt
dir/file1.txt
empty_dir
dir/file1.txt
dir2/file2.txt
file3.txt
file4.txt
file5.txt

Tars:
000001.tar
000002.tar
000003.tar
000004.tar
```

4. `zstash ls --tars -l` after `update` added the `tars` table:
```
$ zstash ls --tars -l --hpss=none
id	name	size	mtime	md5	tar	offset
1	file0.txt	11	2021-04-23 00:30:06	cd62ea5821e9564e12eef26589ad9d05	000000.tar	0	
2	file0_hard.txt	11	2021-04-23 00:30:06	cd62ea5821e9564e12eef26589ad9d05	000000.tar	1024	
3	file0_soft.txt	0	2021-04-23 00:30:06	None	000000.tar	2048	
4	file0_soft_bad.txt	0	2021-04-23 00:30:06	None	000000.tar	2560	
5	file_empty.txt	0	2021-04-23 00:30:06	d41d8cd98f00b204e9800998ecf8427e	000000.tar	3072	
6	dir/file1.txt	11	2021-04-23 00:30:06	ece8b2cf75e0151cbbc849a64c0aeac9	000000.tar	3584	
7	empty_dir	0	2021-04-23 00:30:06	None	000000.tar	4608	
8	dir/file1.txt	24	2021-04-23 00:30:06	dac1d5cf79f0b8977ddb728ab89a4b14	000001.tar	0	
9	dir2/file2.txt	11	2021-04-23 00:30:06	63acce5e0a48a06a8be527a46be35bb7	000001.tar	1024	
10	file3.txt	11	2021-04-23 00:30:07	8402c947a709c83a269055539cf76b62	000002.tar	0	
11	file4.txt	11	2021-04-23 00:30:07	e206616a01ccf3d1718f8e6512bf38dd	000003.tar	0	
12	file5.txt	11	2021-04-23 00:30:08	df03724ee38bd447119e246847830a17	000004.tar	0	

Tars:
id	name	size	md5
1	000001.tar	35	1ee2c4e7cf386e372c1a4014f395eb47	
2	000002.tar	11	60984749dc1aa11989f4d20aae8e1055	
3	000003.tar	11	73d9a05773f176e1e855db1ddfeb9756	
4	000004.tar	11	cc8b7eb9c5754accd5988f1aeb4b586d
```

Notice that `000000.tar` is missing in the last 2 scenarios because that was created on `create` (i.e., before the `tars` table existed).

forsyth2:
Note: it may be a good idea to get some performance benchmarks on time to run with and without the `tars` table

golaz (changes requested):
@forsyth2 : this will be a very nice feature to have! 

I have tested it on NERSC and it (almost) works. But there is a problem with the reported tar file size. See my suggestion for what the problem might be.

Also, with the current implementation, we cannot use the `--no_tars_table` to compare performance with and without the md5 checksums of the tar files, since there are computed in both cases, just not always included in table. I don't think that's a problem. But if you really wanted that option it could be done by not using HashIO when opening the file. Again, I don't know if this is worth the complications.

I ran a test on a medium size archive (~200 GB).

```
# zstash create with md5 of files and tars:
cd /global/cfs/cdirs/e3sm/golaz/E3SM_simulations
time zstash create -v --hpss=none 20200616.v1like.F2010.ne30pg2_r05_oECv3.compy
real	19m49.919s
```

```
# Stand-alone md5 of tar file
time md5sum 000000.tar 
real	8m10.912s
```

The performance looks impressive. 20 minutes to do the whole thing. Just computing the md5 of the tar file alone afterwards takes 8 minutes!

I have some older scripts on NERSC to compare performance of zstash with separate operations (what I call stash). We could modify these tests to do a new performance test including the tar md5. Given the numbers above, the performance gain might be impressive (see /global/homes/g/golaz/E3SM/utils/zstash_performance).

golaz:
@forsyth2 : I had another thought following on my previous comment. With HashIO, we can now keep track of the real size of the tar file as it is built (including metadata). We could use that information to improve the cutoff size. On some occasions, tar were larger than the specified maximum size because metadata was not included in the count. We can now improve that.

Instead of

```
             # Increase tarsize by the size of the current file
            tarsize += size
```

get the real current size

```
            tarsize = tarFileObject.tell()
```

which should make the following more accurate:

```
        if i == nfiles - 1 or tarsize + next_file_size > maxsize:
```

zstash/hpss_utils.py
```diff
@@ -12,7 +12,35 @@
...
-from .settings import BLOCK_SIZE, TupleFilesRowNoId, config, logger
+from .settings import BLOCK_SIZE, TupleFilesRowNoId, TupleTarsRowNoId, config, logger
+from .utils import create_tars_table, tars_table_exists
+
+
+# Minimum output file object
+class HashIO(object):
+    def __init__(self, name: str, mode: str):
+        self.f: _io.BufferedWriter = open(name, mode)
+        self.hash: _hashlib.HASH = hashlib.md5()
```
forsyth2, line 23: Perhaps we could have a `do_hash=True` option.
`self.do_hash = do_hash`

In `write()` and `md5()` update the hash *if* `do_hash` is `True`.

forsyth2:
@golaz I reran the benchmarking with the latest commit ("Second benchmark"). The results I got seem less encouraging.

### Case 1:
With `no_tars_table`: `real 9m50.544s` (590.544s)
md5: `real 6m24.466s` (384.466s)
Total: 975.01s

### Case 2:
Computing the tar md5 automatically: `real 15m37.682s` (937.682s)

### Conclusion
- 1.59x slowdown from not computing the tar's md5 at all
- 1.04x speedup from computing the tar's md5 after running `zstash`.

golaz:
@forsyth2 : I tested your updated code.The size of the tar file is now correct. Thanks for fixing that.

I also ran one timing test. I'm seeing only a small speed advantage (21 vs 22 (13+8) minutes in my test), which is a little surprising to me. Maybe that's all we can expect. Later, we could try tweaking the default block size to see if it has any impact.

One minor detail, do you want to rename the option '--no_tars_table' to '--no_tar_md5' or '--no_tars_md5' since the difference is not just the table any longer. We still don't want to recommend that option, but someone could choose to use it if they really cared about the performance gain from skipping the tar md5.

Go ahead and merge when you are ready.

forsyth2:
@golaz Thank you for reviewing!

You have a 22/21=1.05x speedup, which matches my benchmarking. I found that surprising too.

I changed the option names. It should be noted that `zstash update` will add in the `tars` table if it doesn't exist, so a performance gain from `--no_tars_md5" would only be seen in `zstash create`.

With this new feature merged, we will now be able to increment the minor version in the next release (i.e., 1.0.0 -> 1.1.0).

PR #143, opened 2021-05-12
Update docs for v1.1.0

### LINKED ISSUE #142: Finalize docs for v1.1.0
Before releasing v1.1.0 we should make sure the docs look good. 

E.g.:
- Add information on what's new in v1.1.0.
- The tutorial (https://e3sm-project.github.io/zstash/_build/html/master/tutorial.html) should probably mention what `zstash` version it was run with. (We should also delete line with "INFO: You're using the version at").

forsyth2:
Update docs for v1.1.0. Resolves #142

### DIFF
--- a/docs/source/database.rst
+++ b/docs/source/database.rst
@@ -13,15 +13,14 @@ Structure
 =========
 
 Metadata from zstash archives is stored in a sqlite3 database
-in `zstash/index.db`. The database schema consists of
+in ``<cache>/index.db``, where ``<cache>`` is ``zstash`` by default.
+The database schema consists of
 two tables, one to store configuration parameters (``config``) 
 and the other one for metadata for each archived file (``files``). ::
 
     $ sqlite3 zstash/index.db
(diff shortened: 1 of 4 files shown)


forsyth2:
@golaz I think this is ready for review. I can't think of anything else the docs need before v1.1.0 is released. #140 was the only feature we added for this release and I updated the docs to explain it. I specifically didn't include docs on the `--no_tars_md5` option since that's really only for developers/debuggers.

golaz (approved):
The changes are fine. There are a couple of other places where some content could be added:

- Home page: mention that checksums of the tar files are now computed as well on the fly.
- 'Database' page which lists the sql tables.
- 'Usage - List': should we add one example with --tars option?
-

forsyth2:
@golaz Thanks! I added a second commit to address your suggestions. If those look good, I'll squash the commits and merge.

golaz:
Thanks @forsyth2. Go ahead and merge.

PR #144, opened 2021-05-27
Add Chrysalis HPSS archiving

forsyth2:
Add Chrysalis HPSS archiving

### DIFF
--- a/zstash/create.py
+++ b/zstash/create.py
@@ -14,6 +14,7 @@ from .settings import DEFAULT_CACHE, config, get_db_filename, logger
 from .utils import (
     create_tars_table,
     get_files_to_archive,
+    get_machine,
     run_command,
     tars_table_exists,
 )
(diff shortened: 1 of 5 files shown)


zstash/create.py
```diff
@@ -52,15 +53,22 @@ def create():
...
+        machine: str = get_machine()
+        if machine == "chrysalis":
+            hpss_command = "archive"
+        elif machine == "cori":
+            hpss_command = "hsi -q"
+        else:
+            raise ValueError("Invalid machine={}".format(machine))
         logger.debug("Creating target HPSS directory")
-        mkdir_command: str = "hsi -q mkdir -p {}".format(hpss)
+        mkdir_command: str = "{} mkdir -p {}".format(hpss_command, hpss)
```
forsyth2, line 64: @golaz As far as I can tell, if a file is in `/home/<user>/zstash_archive_test` then `archive` will store it in `/gpfs/fs1/home/<user>/zstash_archive_test/`. I don't see any way to set the HPSS file path to be anything different. This seems like a major obstacle.

zstash/create.py
```diff
@@ -52,15 +53,22 @@ def create():
...
         mkdir_error_str: str = "Could not create HPSS directory: {}".format(hpss)
         run_command(mkdir_command, mkdir_error_str)
 
         # Make sure it is exists and is empty
         logger.debug("Making sure target HPSS directory exists and is empty")
 
-        ls_command: str = 'hsi -q "cd {}; ls -l"'.format(hpss)
+        ls_command: str = '{} "cd {}; ls -l"'.format(hpss_command, hpss)
```
forsyth2, line 71: @golaz `archive` also doesn't appear to be able to take other commands as arguments. Everything has to be done through the limited number of options seen with `archive --help`.

forsyth2:
@rljacob Do you have any more information on the LCRC `archive` tool? It doesn't look like it has the same level of functionality as `hsi` provides -- see my comments above.

forsyth2:
Moving this to the v1.2.0 project board.

We may want to update the documentation (specifically https://e3sm-project.github.io/zstash/_build/html/master/best_practices.html#compy-anvil) to mention that after running `zstash` with `--hpss=none`, users may choose to use the LCRC `archive` tool to store the `zstash` cache/local archive.

rljacob:
I'm told the Chrysalis "archive" tool is a wrapper around this IBM product.  https://www.ibm.com/docs/en/spectrum-protect/8.1.9?topic=commands-archive

zstash/create.py
rljacob: You want the path to be under some generic user or no user at all?

forsyth2:
@chengzhuzhang @rljacob I got an email from LCRC support stating "We've been reengineering our backup environment. After discussing projects needs, we are going to stay with standard backups of E3SM data instead of using an archiving function. These will happen automatically in the background." 

So, I think we can't actually implement Chrysalis HPSS archive functionality. Therefore closing this.

PR #157, opened 2021-07-01
Fixes for rc2

### LINKED ISSUE #151: Update release steps in docs
https://e3sm-project.github.io/zstash/_build/html/master/dev_guide/release.html should be updated to reflect changes in the release process (e.g., from #147).

### LINKED ISSUE #155: Make unit tests independent
Currently, if one unit test fails, it may cause others to fail because a failing test might not return to the correct directory. We'd probably need to edit `tests/base.py`'s `tearDown` method to correct this.

### LINKED ISSUE #156: Mention where `hsi` is available.
Include in the docs that `hsi` is only available on Cori login nodes -- not the (Haswell) compute nodes.

### LINKED ISSUE #159: The python package includes the `tests` directory
The line:
https://github.com/E3SM-Project/zstash/blob/e2b1ab3166e468e7a0813c84a56032e7becbb6ec/setup.py#L9
seems designed to exclude tests but it doesn't actually have the `tests` directory in its excludes.  As a result, it ends up in the conda package and it conflicts with other packages that make the same mistake (e.g. `zppy`).

Instead, it would be much cleaner and safer to do:
```
packages=find_packages(include=['zstash', 'zstash.*'])
```

forsyth2:
Fixes for `rc2`. Currently, this pull request doesn't change any code in `zstash/zstash` so it's not _strictly_ necessary for these changes to be included in `rc2`. The documentation changes should be included in the final release though so that the docs for v1.1.0 are up-to-date.

- Add note in documentation that `hsi` is only available on NERSC log-in nodes. Resolves #156.
- Add potential fix (`shutil._USE_CP_SENDFILE = False`) for Chrysalis `Resource temporarily unavailable` error that appears intermittently when testing using `shutil`.
- Remove updating the `sha` from the release process steps. Resolves #151.
- Make sure unit tests always start from the same directory to ensure a failure in one test doesn't cause directory issues in a later test. Resolves #155.
- Update packages in `setup.py` and change tests to run from `zstash` rather than `zstash/tests`. Resolves #159.


forsyth2:
@tomvothecoder Re: #151, I just removed the steps for updating the `sha` from the release process, since that's no longer necessary. That issue also mentioned changes from #147 -- but it doesn't look like anything in that PR would require a change to the release process docs.

tomvothecoder:
@forsyth2 Cool, thanks for doing that. I removed that section in #158, but it is intended for `v1.2.0`.

In #158, I added these bullets about the Anaconda label assignment:
https://github.com/E3SM-Project/zstash/pull/158/files#r662485876

forsyth2:
@tomvothecoder Thanks, I just added those bullets (https://github.com/E3SM-Project/zstash/pull/157/files#diff-e0e89cf3a4539a9f8238ada29d453c48b19dc8029153f6877e206c39c61fd747R118-R120) so they can be part of the v1.1.0 docs.

PR #170, opened 2021-11-29
Add option to specify tars to check or extract

### LINKED ISSUE #164: specify the starting tar file for checking or parallel checking
Is it possible to specify which tar file to start the checking? Or checking multiple tar files simultaneously.

For large production runs the tar file size is huge (almost 150 TB for the 1000-year spin-up E3SMv2 NARRM run). Checking the tar files one by one can take a really long time. NERSC just had a disruption on hsi and failed my checking job. I had to recheck the files. In this case, it would be really helpful to be able to specify which tar file to start with to avoid duplicated work.

As far as I know, the zstash check job is done on 1 tar file at a time. The performance can be significantly improved if checking multiple files at once. The HPSS system normally allows retrieving multiple files simultaneously.

forsyth2:
Specify starting tar. Resolves #164.

### DIFF
--- a/zstash/extract.py
+++ b/zstash/extract.py
@@ -7,6 +7,7 @@ import heapq
 import logging
 import multiprocessing
 import os.path
+import re
 import sqlite3
 import sys
 import tarfile
@@ -85,6 +86,7 @@ def setup_extract() -> Tuple[argparse.Namespace, str]:
         type=str,
         help='path to the zstash archive on the local file system. The default name is "zstash".',
     )
+    optional.add_argument("--tars", type=str, help="specify which tars to process")
     optional.add_argument(
         "-v", "--verbose", action="store_true", help="increase output verbosity"
     )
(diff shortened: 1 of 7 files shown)


zstash/extract.py
```diff
@@ -162,14 +189,30 @@ def extract_database(
...
+            )
+            matches_ = matches_ + cur.fetchall()
+    else:  # TODO: should this always be run (i.e., not if/else?)
+        # Find matching files
+        for args_file in args.files:
+            cur.execute(
+                u"select * from files where name GLOB ? or tar GLOB ?",
+                (args_file, args_file),
+            )
+            matches_ = matches_ + cur.fetchall()
```
forsyth2, line 215: Should we still run this section if `--tars` is being used? That is, is there a use case where a user would specify `zstash check --hpss=<HPSS> --tars=<tars> <specific files>`? It seems like if a user is requesting specific files, it wouldn't be necessary to specify which tars to look at.

forsyth2 (commented):
@golaz This is ready for review.

zstash/extract.py
tangq: That's right. When specifying the tar files, we don't need the `--tars` option - It's redundant.

tangq (approved):
Looks good by visual check. If testing successfully and removing the redundant `--tars` option when specifying the file names to check, it can be merged.

zstash/extract.py
forsyth2: @tangq If a user specifies both `<tars>` and `<specific files>`, what should be the expected behavior?
1. Just check/extract `<tars>`
2. Just check/extract `<specific files>`
3. Print error message saying to only specify one

tangq: I'd say option 3. What do you think?

forsyth2: I would agree. @golaz which is your preference?

golaz: @forsyth2 and @tangq : this is an edge case and I would think that it is unlikely a user would want to specify `<tars>` and `<specific files>` simultaneously. Therefore, if it's relatively straightforward to implement, I would suggest option 3 as well to make sure there is no room for misunderstanding from an unsuspecting user.

golaz (approved):
@forsyth2 : the code looks good. I made a few minor suggestions. Let me know if you'd like me to try it out myself.

forsyth2:
@golaz @tangq I created a commit addressing your comments: https://github.com/E3SM-Project/zstash/pull/170/commits/e139de7d852ca72159bce37e0b0a7f66f47de618

Let me know if that looks good. If so, I'll squash and merge this pull request.

golaz:
Thanks @forsyth2. The changes look good to me.

Issue #176, opened 2022-01-10
Fix `chgrp` test failure


forsyth2:
`testChgrpHPSS` fails with `AssertionError: Command=`hsi ls -l zstash_test`. Errors=['This was supposed to be found, but was not: 3.', 'This was supposed to be found, but was not: m.']`. This occurs as far as back as v1.0.0 -- I'm unsure why this is happening, since the unit tests have been run multiple time since then.

In `tests/test_chgrp.py`, `expected_present` is a string, but should be a list. 
```
                expected_present = 'e3sm'
            else:
                expected_present = 'chgrp: HPSS is unavailable'
            self.check_strings(cmd, output + err, expected_present, ['ERROR'])
```
Note, however, that even changing it to a list does not make the test pass.

forsyth2:
Partially fixed by #179.

PR #177, opened 2022-01-10
Remove keep from config

### LINKED ISSUE #63: Remove keep from config
Remove `keep` from the `config`. Doing this will mean tar files won't be kept or removed based on using the `keep` option in an earlier command. (Although it doesn't seem like this is the current behavior anyway).

forsyth2:
Remove `keep` from the `config`. Resolves #63


forsyth2:
@golaz This is ready for review. We're only trying to remove `keep` from the config, correct? That is, we're keeping the `--keep` option available?

The unit tests still pass (aside from the unrelated failure caused by #176).

golaz (approved):
Thanks for making the change. Looks like it actually simplifies the code slightly.

Issue #178, opened 2022-01-11
test_globus fails on CI/CD only


forsyth2:
`test_globus` fails on CI/CD only. 

From #154:

tests/test_globus.py requires:

- a Globus endpoint on a local machine,
- ~/.zstash.ini with an UUID of the local Globus endpoint set,
- Globus access tokens

to succeed. It can be run as a part of the CI/CD build workflow but it would some extra setup on a CI/CD machine. The test can be run easily on any E3SM machine.

forsyth2:
Partially fixed by #179.

PR #179, opened 2022-01-11
Fix tests

forsyth2:
- Partial fix to #176 (disable part of the `chgrp` HPSS test rather than getting past the "Must run interactively to update .netrc" error)
- Partial fix to #178 (skip the test on CI/CD rather than setting local endpoint)

### DIFF
--- a/tests/test_chgrp.py
+++ b/tests/test_chgrp.py
@@ -31,12 +31,14 @@ class TestChgrp(TestZstash):
             output, err = run_cmd(cmd)
             if use_hpss:
                 self.check_strings(cmd, output + err, [], ["ERROR"])
-                print("Now check that the files are in the {} group".format(GROUP))
-                cmd = "hsi ls -l {}".format(self.hpss_path)
-                output, err = run_cmd(cmd)
-                expected_present = "e3sm"
+                # Ignore this part due to "Must run interactively to update .netrc" error
+                # print("Now check that the files are in the {} group".format(GROUP))
+                # cmd = "hsi ls -l {}".format(self.hpss_path)
+                # output, err = run_cmd(cmd)
+                # expected_present = ["e3sm"]
+                expected_present = []
             else:
-                expected_present = "chgrp: HPSS is unavailable"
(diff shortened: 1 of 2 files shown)


tests/test_chgrp.py
```diff
@@ -31,12 +31,14 @@ def helperChgrp(self, test_name, hpss_path, zstash_path=ZSTASH_PATH):
...
-                print("Now check that the files are in the {} group".format(GROUP))
-                cmd = "hsi ls -l {}".format(self.hpss_path)
-                output, err = run_cmd(cmd)
-                expected_present = "e3sm"
+                # Ignore this part due to "Must run interactively to update .netrc" error
+                # print("Now check that the files are in the {} group".format(GROUP))
+                # cmd = "hsi ls -l {}".format(self.hpss_path)
+                # output, err = run_cmd(cmd)
+                # expected_present = ["e3sm"]
+                expected_present = []
```
forsyth2, line 39: Something must have changed with NERSC. Running `hsi ls -l` now gives a "Must run interactively to update .netrc" error.

I followed all the relevant steps from the test manually and was able to run `hsi ls -l` though, so it must be an issue with the way the tests are run.

tests/test_globus.py
```diff
@@ -49,7 +49,10 @@ def preactivate_globus(self):
                     local_endpoint = regex_endpoint_map.get(pattern)
                     break
         if not local_endpoint:
-            self.fail("{} does not have the local Globus endpoint set".format(ini_path))
+            # self.fail("{} does not have the local Globus endpoint set".format(ini_path))
+            self.skipTest(
+                "{} does not have the local Globus endpoint set".format(ini_path)
+            )
```
forsyth2, line 55: This test works fine when run by a developer directly, so I'm just skipping it on CI/CD for now.

forsyth2 (commented):
@golaz This is ready for review. Let me know if we should try to implement more-involved solutions instead.

golaz (approved):

PR #181, opened 2022-01-14
Improve docs on check

### LINKED ISSUE #180: Update usage docs for Check
Clarify how `check` works when `--hpss=none` on https://e3sm-project.github.io/zstash/_build/html/master/usage.html#check

forsyth2:
Improve docs on `check`. Resolves #180.

### DIFF
--- a/docs/source/usage.rst
+++ b/docs/source/usage.rst
@@ -132,7 +132,8 @@ go to a **new, empty directory** and run: ::
 
 where
 
-* ``--hpss=<path to HPSS>`` specifies the destination path on the HPSS file system.
+* ``--hpss=<path to HPSS>`` specifies the destination path on the HPSS file system. If ``--hpss=none``,
+  then ``zstash check`` will check the cache.
 * ``--workers=<num of processes>`` an optional argument which specifies the number of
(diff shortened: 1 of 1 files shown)


forsyth2:
@chengzhuzhang Does this usage example look good?

chengzhuzhang (commented):
Looks good. The new use case doesn't require `--keep` right? Just want to make sure that the tar balls won't be purged.

forsyth2:
> The new use case doesn't require --keep right?

No, when `--hpss=none`, `--keep` is essentially applied automatically.

chengzhuzhang (commented):
Thanks for clarifying. In this case, you may want to add this in `--keep` to explain?

forsyth2:
https://e3sm-project.github.io/zstash/_build/html/master/tutorial.html#keep-files does mention "If `--keep` is used with `--hpss=none`, there won’t be a noticeable effect since the latter keeps tar files by default."

I could perhaps add a note at the top of the Usage docs highlighting that though.

forsyth2:
> I could perhaps add a note at the top of the Usage docs highlighting that though.

I added this in #182.

PR #185, opened 2022-01-17
Activate Globus endpoints before creating tarballs

lukaszlacinski:
Zstash commands `create` and `update` create a tarball and then transfer the tarball to HPSS using hsi or Globus. To transfer the tarball to a remote Globus endpoint, zstash needs a valid Globus OAuth2 access token and both local and remote Globus endpoints must be activated. If one of the conditions is not fulfilled, zstash will fails after spending even several minutes on creating the tarball. This PR moves the Globus initialization steps:

- getting an access token from the Globus Auth service,
- activating both remote and local Globus endpoints

before zstash creates a tarball.

### DIFF
--- a/zstash/create.py
+++ b/zstash/create.py
@@ -10,6 +10,7 @@ from typing import Any, List, Tuple
 
 from six.moves.urllib.parse import urlparse
 
+from .globus import globus_activate
 from .hpss import hpss_put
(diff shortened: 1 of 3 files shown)


forsyth2:
Completed as part of #214.

Issue #217, opened 2022-08-26
"file could not be opened successfully" on a v2.LR.historical_0101 tar ball


chengzhuzhang:
I'm trying to extract the v2.LR.historical_0101 run from NERSC HPSS archive and ran into "file could not be opened successfully" error on zstash/00005.tar when beginning extract `archive/atm/hist/v2.LR.historical_0101.eam.h0.1978-05.nc`. 

To reproduce:

```
source /global/common/software/e3sm/anaconda_envs/load_latest_e3sm_unified_cori-haswell.sh
zstash extract  --hpss=/home/f/forsyth/E3SMv2/v2.LR.historical_0101 "archive/atm/hist/*h0.197[8-9]*"
```
Not sure how to trouble shooting, maybe it's a problem in the archive?

The Traceback is as follows:

```
INFO: Opening tar archive zstash/000005.tar
Traceback (most recent call last):
  File "/global/common/software/e3sm/anaconda_envs/base/envs/e3sm_unified_1.7.0_nompi/bin/zstash", line 10, in <module>
    sys.exit(main())
  File "/global/common/software/e3sm/anaconda_envs/base/envs/e3sm_unified_1.7.0_nompi/lib/python3.9/site-packages/zstash/main.py", line 64, in main
    extract()
  File "/global/common/software/e3sm/anaconda_envs/base/envs/e3sm_unified_1.7.0_nompi/lib/python3.9/site-packages/zstash/extract.py", line 45, in extract
    failures: List[FilesRow] = extract_database(args, cache, keep_files)
  File "/global/common/software/e3sm/anaconda_envs/base/envs/e3sm_unified_1.7.0_nompi/lib/python3.9/site-packages/zstash/extract.py", line 275, in extract_database
    failures = extractFiles(matches, keep_files, keep, cache)
  File "/global/common/software/e3sm/anaconda_envs/base/envs/e3sm_unified_1.7.0_nompi/lib/python3.9/site-packages/zstash/extract.py", line 444, in extractFiles
    tar: tarfile.TarFile = tarfile.open(tfname, "r")
  File "/global/common/software/e3sm/anaconda_envs/base/envs/e3sm_unified_1.7.0_nompi/lib/python3.9/tarfile.py", line 1625, in open
    raise ReadError("file could not be opened successfully")

```

chengzhuzhang:
[...]

Issue #212, opened 2022-06-21
Handle incomplete tar files


forsyth2:
Handle incomplete tar files. #198 refers to simply retrying after a `hsi get` failure. This issue is for handling an incomplete `tar` file, which may result from a `hsi get` failure.

forsyth2:
@golaz FYI ^

Feel free to add any further comments/explanation.

forsyth2:
It looks like we do have information on the sizes of tars (https://github.com/E3SM-Project/zstash/blob/master/zstash/settings.py#L72), so that may be useful.

PR #218, opened 2022-08-29
hsi get retry

### LINKED ISSUE #198: Option to retry after hsi get failure
Retrieval of tar files from HPSS on NERSC has been less reliable lately. `hsi get` occasionally fails for unknown reasons, but retrying works. We should consider adding an option to retry after a failed hsi operation.

forsyth2:
`hsi get` retry -- resolves #198

### DIFF
--- a/docs/source/usage.rst
+++ b/docs/source/usage.rst
@@ -304,6 +304,10 @@ where
 * ``--keep`` to keep a copy of the tar files on the local file system after
   they have been extracted from the archive. Normally, they are deleted after
   successful transfer.
+* ``--retries`` to set the number of times to retry ``hsi get`` if it is unsuccessful.
+  The default is 1 retry (2 tries total). Note: for a retry to occur automatically because of
+  an incomplete tar file, then the archive you're extracting from
+  must have been created using ``zstash >= v1.1.0``.
 * ``--tars`` to	specify	specific tars to extract. See "Check" above for example usage.
 * ``-v`` increases output verbosity.
 * ``[files]`` is a list of files to be extracted (standard wildcards supported).
--- a/zstash/extract.py
+++ b/zstash/extract.py
@@ -30,7 +30,7 @@ from .settings import (
     get_db_filename,
     logger,
 )
-from .utils import update_config
+from .utils import tars_table_exists, update_config
 
 
 def extract(keep_files: bool = True):
@@ -93,6 +93,9 @@ def setup_extract() -> Tuple[argparse.Namespace, str]:
         "--cache",
         type=str,
         help='path to the zstash archive on the local file system. The default name is "zstash".',
+    )
+    optional.add_argument(
+        "--retries", type=int, default=1, help="number of times to retry an hsi command"
     )
     optional.add_argument("--tars", type=str, help="specify which tars to process")
     optional.add_argument(
@@ -270,9 +273,11 @@ def extract_database(
     failures: List[FilesRow]
     if args.workers > 1:
         logger.debug("Running zstash {} with multiprocessing".format(cmd))
-        failures = multiprocess_extract(args.workers, matches, keep_files, keep, cache)
+        failures = multiprocess_extract(
+            args.workers, matches, keep_files, keep, cache, cur, args
+        )
     else:
-        failures = extractFiles(matches, keep_files, keep, cache)
+        failures = extractFiles(matches, keep_files, keep, cache, cur, args)
 
     # Close database
     logger.debug("Closing index database")
@@ -287,6 +292,8 @@ def multiprocess_extract(
     keep_files: bool,
     keep_tars: Optional[bool],
     cache: str,
+    cur: sqlite3.Cursor,
+    args: argparse.Namespace,
 ) -> List[FilesRow]:
     """
     Extract the files from the matches in parallel.
@@ -359,7 +366,7 @@ def multiprocess_extract(
         )
         process: multiprocessing.Process = multiprocessing.Process(
             target=extractFiles,
-            args=(matches, keep_files, keep_tars, cache, worker),
+            args=(matches, keep_files, keep_tars, cache, cur, args, worker),
             daemon=True,
         )
         process.start()
@@ -379,12 +386,36 @@ def multiprocess_extract(
     return failures
 
 
+def check_sizes_match(cur, tfname):
+    match: bool
+    if cur and tars_table_exists(cur):
+        logger.info(f"{tfname} exists. Checking expected size matches actual size.")
+        actual_size = os.path.getsize(tfname)
+        name_only = os.path.split(tfname)[1]
+        cur.execute(f"select size from tars where name is '{name_only}';")
+        expected_size: int = cur.fetchall()[0][0]
+        if expected_size != actual_size:
+            logger.info(
+                f"{name_only}: expected size={expected_size} != {actual_size}=actual_size"
+            )
+            match = False
+        else:
+            # Sizes match
+            match = True
+    else:
+        # Cannot access size information; assume the sizes match.
+        match = True
+    return match
+
+
 # FIXME: C901 'extractFiles' is too complex (33)
 def extractFiles(  # noqa: C901
     files: List[FilesRow],
     keep_files: bool,
     keep_tars: Optional[bool],
     cache: str,
+    cur: sqlite3.Cursor,
+    args: argparse.Namespace,
     multiprocess_worker: Optional[parallel.ExtractWorker] = None,
 ) -> List[FilesRow]:
     """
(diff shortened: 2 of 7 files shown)


zstash/hpss.py
```diff
@@ -95,7 +95,14 @@ def hpss_transfer(
             # Transfer file using `hsi`
             command: str = 'hsi -q "cd {}; {} {}"'.format(hpss, transfer_command, name)
             error_str: str = "Transferring file {} HPSS: {}".format(transfer_word, name)
-            run_command(command, error_str)
+            if transfer_type == "get":
```
forsyth2, line 98: Do we want to retry any failing `hsi` operation, or only `hsi get`?

zstash/hpss.py
```diff
@@ -95,7 +95,14 @@ def hpss_transfer(
...
+                try:
+                    run_command(command, error_str)
+                except RuntimeError:
+                    # Retry if `hsi get` fails.
+                    run_command(command, error_str)
```
forsyth2, line 103: Does it make sense to always retry `hsi` get` in the event of failure? Or do we indeed want a command line option to request this?

forsyth2 (commented):
@chengzhuzhang @golaz Luckily this issue doesn't require too many code changes, but I want to confirm a couple design choices -- see comments. Thanks!

forsyth2:
#212 specifically addresses the incomplete tar error, whereas #198 is to simply retry `hsi get`. #217 occurred because of #212. This pull request should resolve all three issues.

We can add an additional test for the tar file being on disk (for backwards-compatibility, run check if tars table exists). We could make a command line option for number of retry attempts (default=1).

zstash/hpss.py
```diff
@@ -95,7 +98,28 @@ def hpss_transfer(
...
+                        and cur
+                        and tars_table_exists(cur)
+                    ):
+                        cur.execute(f"select size from tars where name is '{name}';")
+                        expected_size: int = cur.fetchall()[0][0]
+                        if expected_size > actual_size:
+                            # Rerun since tar was incomplete
+                            # Break out of the for-loop and run the try-except block again
+                            continue
+                    raise e
```
forsyth2, line 122: @golaz @chengzhuzhang Here is the primary piece of code for this pull request. It's very challenging to test behavior that only occurs when there is an error, but I believe I have this set up correctly. Please let me know if this looks good to you as well. Thanks! (I did add a unit test, but that doesn't create an error to test the `except` block.)

docs/source/usage.rst
```diff
@@ -304,6 +304,9 @@ where
 * ``--keep`` to keep a copy of the tar files on the local file system after
   they have been extracted from the archive. Normally, they are deleted after
   successful transfer.
+* ``--retries`` to set the number of times to retry ``hsi get`` if a tar file is incomplete.
+  The default is 1 retry (2 tries total). Note that the archive you're extracting from
+  must have been created using ``zstash >= v1.1.0`` for this to work.
```
forsyth2, line 309: @golaz @chengzhuzhang Here is an explanation of the `--retries` option.

zstash/hpss.py
```diff
@@ -95,7 +98,28 @@ def hpss_transfer(
...
+                        (tries > 0)
+                        and (transfer_type == "get")
+                        and cur
+                        and tars_table_exists(cur)
+                    ):
+                        cur.execute(f"select size from tars where name is '{name}';")
+                        expected_size: int = cur.fetchall()[0][0]
+                        if expected_size > actual_size:
+                            # Rerun since tar was incomplete
+                            # Break out of the for-loop and run the try-except block again
```
forsyth2, line 120: Note: this comment is from an earlier iteration of work; this should be changed to simply "Run the try-except block again"

golaz (changes requested):
The code needs to be modified in two places to achieve the desired goals:

- When deciding whether the tar file needs to be retrieved from tape, check its size (if zstash database has the information).
- Retry if hsi get returns a failure. Regardless of file size and zstash version.

zstash/hpss.py
forsyth2: @golaz re: `check the tar file size regardless of hsi get status` -- Sure, I can copy the `if expected_size > actual_size:` logic to the `try` block so that it is checked on both success and failure.

You also mentioned the logic should be at https://github.com/E3SM-Project/zstash/blob/main/zstash/extract.py#L435 (`if not os.path.exists(tfname):`) instead, but I'm not seeing what effect that has. If there's no `tfname`, then we have nothing to get a size from -- so I don't see what logic we would apply at this level. As for the case where `tfname` does exist, we have: `tfname` -> `file_path` in `hpss_get` -> `file_path` in `hpss_transfer` -> `path, name = os.path.split(file_path)` -> `actual_size = os.path.getsize(name)`.

golaz: @forsyth2, regarding the check around https://github.com/E3SM-Project/zstash/blob/753452869a27e683b515fc285a5869fa304fdc40/zstash/extract.py#L435

I would modify to do the following: retrieve the tar from from HPSS 
- if it is not in the disk cache (what we do now)
- OR if the file is on disk cache, but its size does not match the entry in the database (assuming version of zstash archive has that information).

This would help recover from previous errors.

zstash/extract.py
```diff
@@ -432,13 +441,47 @@ def extractFiles(  # noqa: C901
...
-                    hpss: str = config.hpss
-                else:
-                    raise TypeError("Invalid config.hpss={}".format(config.hpss))
-                hpss_get(hpss, tfname, cache)
+            if config.hpss is not None:
+                hpss: str = config.hpss
+            else:
+                raise TypeError("Invalid config.hpss={}".format(config.hpss))
+            tries = args.retries + 1
+            # Set to True to test the `--retries` option with a forced failure.
```
forsyth2, line 449: This is just a manual test that can be run to double-check the logic.

zstash/extract.py
```diff
@@ -432,13 +441,47 @@ def extractFiles(  # noqa: C901
...
+                        test_retry = False
+                        raise RuntimeError
+                    if not os.path.exists(tfname):
+                        # Will need to retrieve from HPSS
+                        hpss_get(hpss, tfname, cache)
+                    elif cur and tars_table_exists(cur):
+                        logger.info(
+                            f"{tfname} exists. Checking expected size matches actual size."
+                        )
+                        actual_size = os.path.getsize(tfname)
```
forsyth2, line 465: I did a test by doing the following:

Changed this code block to be:
```
                    if not os.path.exists(tfname):
                        # Will need to retrieve from HPSS
                        hpss_get(hpss, tfname, cache)
                        raise RuntimeError # NEW -- on retry, tfname will exist and logic will go into elif block
                    elif cur and tars_table_exists(cur):
                        logger.info(
                            f"{tfname} exists. Checking expected size matches actual size."
                        )
                        actual_size = os.path.getsize("/global/homes/f/forsyth/zstash_test/add_files.sh") # CHANGED -- this is a much smaller file, so the sizes won't match
```
and run a modified version of the tutorial code:
```
$ emacs setup.sh
mkdir zstash_demo
mkdir zstash_demo/empty_dir
mkdir zstash_demo/dir
echo 'file0 stuff' > zstash_demo/file0.txt
echo '' > zstash_demo/file_empty.txt
echo 'file1 stuff' > zstash_demo/dir/file1.txt
$ emacs add_files.sh
mkdir zstash_demo/dir2
echo 'file2 stuff' > zstash_demo/dir2/file2.txt
echo 'file1 stuff with changes' > zstash_demo/dir/file1.txt
$ chmod 700 setup.sh
$ chmod 700 add_files.sh

$ ./setup.sh
$ zstash create --hpss=zstash_retries zstash_demo
$ ./add_files.sh
$ cd zstash_demo
$ zstash update --hpss=zstash_retries
$ cd ..
$ mkdir zstash_extraction2 && cd zstash_extraction2
$ zstash extract --hpss=zstash_retries --retries=3
```

forsyth2 (commented):
@golaz This is ready for another review. Thanks!

zstash/extract.py
```diff
@@ -432,13 +441,47 @@ def extractFiles(  # noqa: C901
...
+                        name_only = os.path.split(tfname)[1]
+                        cur.execute(
+                            f"select size from tars where name is '{name_only}';"
+                        )
+                        expected_size: int = cur.fetchall()[0][0]
+                        if expected_size > actual_size:
+                            logger.info(
+                                f"{name_only}: expected size={expected_size} > {actual_size}=actual_size"
+                            )
+                            hpss_get(hpss, tfname, cache)
```
forsyth2, line 475: If there's no `tars_table`, if the previous `hpss_get` had an exception, it won't retry.

zstash/extract.py
```diff
@@ -432,13 +441,47 @@ def extractFiles(  # noqa: C901
...
+                        logger.info(
+                            f"{tfname} exists. Checking expected size matches actual size."
+                        )
+                        actual_size = os.path.getsize(tfname)
+                        name_only = os.path.split(tfname)[1]
+                        cur.execute(
+                            f"select size from tars where name is '{name_only}';"
+                        )
+                        expected_size: int = cur.fetchall()[0][0]
+                        if expected_size > actual_size:
```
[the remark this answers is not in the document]
forsyth2: - 2x2 check -- file exists & sizes match up.
  - T & N/A: skip `hpss_get`
  - T & T: skip `hpss_get`
  - T & F: run
  - F & N/A: run
  - F & T: run
  - F & F: N/A -- can't compare sizes.
- Redo size check *after* `hpss_get`

Perhaps have a flag variable `should_retrieve` and then at the end call `hpss_get`.

Nest logic:
```
if file doesn't exist:
  do_retrieve = True # F -- all 3 cases
elif 
  if we can check the size:
    if sizes match:
      do_retrieve = False # T & T
   else:
      do_retrieve = True # T & F
 else:
   do_retrieve = False # T & N/A
if do_retrieve:
  hpss_get
  if we can check the size AND the sizes don't match:
    raise exception ourselves
```

forsyth2:
@golaz Updated retry logic in the latest commit -- https://github.com/E3SM-Project/zstash/pull/218/commits/4b0a29579306c0c874fd17a6a4422d1bfe67761f.

If that looks good, can you approve the PR? Then, I can merge. Thanks!

golaz (approved):
@forsyth2: this looks good to me now. I like your function to verify the size of the tar files. It makes for for a simpler and cleaner implementation of the various checks.

forsyth2:
Thanks @golaz!

PR #226, opened 2022-10-20
update e3sm U paths

chengzhuzhang:
Update both acme1 and andes paths for activating e3sm_unified.


forsyth2 (approved):
@chengzhuzhang Thanks!

Issue #243, opened 2022-12-08
Better support Globus on compute nodes


forsyth2:
As discussed in #241, 'add r"b\d+.lcrc.anl.gov" to globus.py to match the Blues compute nodes'.

forsyth2:
Resolved by #245.

PR #245, opened 2022-12-21
Add regex matching Blues compute nodes to Globus endpoints mapping

lukaszlacinski:
It solves https://github.com/E3SM-Project/zstash/issues/243 that was discussed here https://github.com/E3SM-Project/zstash/discussions/241.


forsyth2:
Thanks @lukaszlacinski!

I can't get an interactive node on Blues to test this, but it looks like a simple enough change. Merging.

PR #246, opened 2023-01-09
Detect Perlmutter nodes and set corresponding Globus endpoint

lukaszlacinski:
NERSC Perlmutter nodes do not have an FQDN set properly. It means that `socket.getfqdn()` and regex cannot be used to determine what machine zstash is run on and set a corresponding Globus endpoint UUID. This PR adds another detection method that relies on `$NERSC_HOST` environment variable that is set on NERSC machines. If the variable is set to `"Perlmutter"` or "unknown", a Perlmutter Globus endpoint will be used for the machine zstash executed on.

### DIFF
--- a/zstash/globus.py
+++ b/zstash/globus.py
@@ -70,6 +70,13 @@ def globus_activate(hpss: str):
             if re.fullmatch(pattern, fqdn):
                 local_endpoint = regex_endpoint_map.get(pattern)
                 break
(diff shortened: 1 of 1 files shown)


forsyth2:
Thanks @lukaszlacinski for implementing this. The unit tests pass on Perlmutter. Merging.

PR #264, opened 2023-05-16
--include option

### LINKED ISSUE #199: Add --include option
In some cases, it would be useful to have an `--include` option in addition to the existing `--exclude` option. If both options are specified at the same time, start with all files in `--include` and remove files as specified in `--exclude`.

forsyth2:
Add `--include` option. Resolves #199.

### DIFF
--- a/zstash/utils.py
+++ b/zstash/utils.py
@@ -10,29 +10,40 @@ from typing import Any, List, Tuple
 from .settings import TupleTarsRow, config, logger
 
 
-def exclude_files(exclude: str, files: List[str]) -> List[str]:
+def filter_files(subset: str, files: List[str], include: bool) -> List[str]:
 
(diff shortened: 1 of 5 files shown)


zstash/utils.py
```diff
@@ -82,7 +93,11 @@ def get_files_to_archive(cache: str, exclude: str) -> List[str]:
         if x[0] != os.path.join(".", cache)
     ]
 
-    # Eliminate files based on exclude pattern
+    # First, add files based on include pattern
+    if include is not None:
+        files = include_files(include, files)
```
forsyth2, line 98: It is sufficient to determine which files to add to the `zstash archive` after getting the entire list of files, as the archiving step (which comes later) is the computationally expensive part.

PR #186, opened 2022-01-18
Fix conda requirements

forsyth2:
Fix conda requirements causing https://github.com/E3SM-Project/zstash/releases/tag/v1.2.0rc1 to not build properly.

PR #154, opened 2021-06-18
Extend --hpss to accept a Globus URL

lukaszlacinski:
This PR adds Globus support. With this PR, the zstash tool can work with any HPSS storage that is accessible as a Globus endpoint. The --hpss option recognizes a Globus URL:

`--hpss=globus://<GLOBUS_ENDPOINT_UUID><PATH>`

Two symbolic names, "alcf" and "nersc", are also recognized as GLOBUS_ENDPOINT_UUID and substituted internally with a UUID corresponding to ALCF HPSS or NERSC HPSS Globus endpoints.
To transfer files using Globus Transfer Service, zstash must know a UUID of a local Globus endpoint. Zstash detects login nodes of four supercomputers: ALCF Theta, LCRC Blues, LCRC Chrysalis, NERSC Cori, and assumes a local endpoint UUID accordingly. On other machines, zstash requires a local Globus endpoint UUID to be set in ~/.zstash.ini.

PR #271, opened 2023-06-15
Change workflow to mamba

### LINKED ISSUE #252: Change workflow from `pip` to `mamba/boa`
Change workflow from `pip` to `mamba/boa`

forsyth2:
Change workflow to mamba. Resolves #252. Equivalent of `zppy`'s https://github.com/E3SM-Project/zppy/pull/429 & https://github.com/E3SM-Project/zppy/pull/428

### DIFF
--- a/setup.py
+++ b/setup.py
@@ -8,10 +8,5 @@ setup(
     description="Long term HPSS archiving software for E3SM",
     packages=find_packages(include=["zstash", "zstash.*"]),
     python_requires=">=3.6",
-    install_requires=[
-        "fair-research-login>=0.2.6,<0.3.0",
-        "globus-sdk>=3.0.0,<4.0.0",
-        "six",
-    ],
     entry_points={"console_scripts": ["zstash=zstash.main:main"]},
 )
--- a/.github/workflows/build_workflow.yml
+++ b/.github/workflows/build_workflow.yml
@@ -44,7 +44,6 @@ jobs:
           key: ${{ runner.os }}-conda-${{ env.CACHE_NUMBER }}-${{
             hashFiles('conda/dev.yml') }}
 
-
       - name: Build Conda Environment
         uses: conda-incubator/setup-miniconda@v2
         with:
@@ -56,8 +55,6 @@ jobs:
           environment-file: conda/dev.yml
           channel-priority: strict
           auto-update-conda: true
-          # IMPORTANT: This needs to be set for caching to work properly!
-          use-only-tar-bz2: true
 
       - name: Install `zstash` Package
         run: |
@@ -82,25 +79,26 @@ jobs:
           persist-credentials: false
           fetch-depth: 0
 
-      - name: Set up Python
-        uses: actions/setup-python@v4
-        with:
-          python-version: 3.9
-
-      - name: Cache pip
+      - name: Cache Conda
         uses: actions/cache@v3
+        env:
+          CACHE_NUMBER: 0
         with:
-          # This path is specific to Ubuntu
-          path: ~/.cache/pip
-          # Look to see if there is a cache hit for the corresponding requirements file
-          key: ${{ runner.os }}-pip-publish-docs
-          restore-keys: |
-            ${{ runner.os }}-pip-
-            ${{ runner.os }}-
+          path: ~/conda_pkgs_dir
+          key: ${{ runner.os }}-conda-${{ env.CACHE_NUMBER }}-${{
+            hashFiles('conda/dev.yml') }}
 
-      - name: Install Dependencies
-        run: |
-          mamba install -y sphinx==5.2.3 sphinx_rtd_theme==1.0.0 sphinx-multiversion==0.2.4 docutils==0.16 "jinja2<3.1"
+      - name: Build Conda Environment
+        uses: conda-incubator/setup-miniconda@v2
+        with:
+          activate-environment: zstash_dev
+          miniforge-variant: Mambaforge
+          miniforge-version: latest
+          use-mamba: true
+          mamba-version: "*"
+          environment-file: conda/dev.yml
+          channel-priority: strict
+          auto-update-conda: true
 
       - name: Build Sphinx Docs
         run: |
--- a/.github/workflows/build_workflow.yml
+++ b/.github/workflows/build_workflow.yml
@@ -35,10 +35,29 @@ jobs:
     steps:
       - uses: actions/checkout@v3
 
-      - name: Set up Python
-        uses: actions/setup-python@v4
+      - name: Cache Conda
+        uses: actions/cache@v3
+        env:
+          CACHE_NUMBER: 0
         with:
-          python-version: 3.9
+          path: ~/conda_pkgs_dir
+          key: ${{ runner.os }}-conda-${{ env.CACHE_NUMBER }}-${{
+            hashFiles('conda/dev.yml') }}
+
+
+      - name: Build Conda Environment
+        uses: conda-incubator/setup-miniconda@v2
+        with:
+          activate-environment: zstash_dev
+          miniforge-variant: Mambaforge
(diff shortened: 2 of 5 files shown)


forsyth2 (commented):
@xylar @tomvothecoder I made effectively the same changes as in https://github.com/E3SM-Project/zppy/pull/429 & https://github.com/E3SM-Project/zppy/pull/428. Does that seem fine?

Remaining instances of `pip`, for reference:
```
$ git grep -n pip ':!docs/_build_old/*'
.github/workflows/build_workflow.yml:45:          python -m pip install --upgrade pip
.github/workflows/build_workflow.yml:46:          pip install .
.github/workflows/build_workflow.yml:71:      - name: Cache pip
.github/workflows/build_workflow.yml:75:          path: ~/.cache/pip
.github/workflows/build_workflow.yml:77:          key: ${{ runner.os }}-pip-publish-docs
.github/workflows/build_workflow.yml:79:            ${{ runner.os }}-pip-
.github/workflows/release_workflow.yml:25:      - name: Cache pip
.github/workflows/release_workflow.yml:29:          path: ~/.cache/pip
.github/workflows/release_workflow.yml:31:          key: ${{ runner.os }}-pip-publish-docs
.github/workflows/release_workflow.yml:33:            ${{ runner.os }}-pip-
conda/dev.yml:8:  - pip=22.2.2
conda/meta.yaml:14:  script: "{{ PYTHON }} -m pip install . --no-deps -vv"
conda/meta.yaml:20:    - pip
docs/source/getting_started.rst:184:Instead, the developer will ``pip install .`` to build ``zstash`` with changes
docs/source/getting_started.rst:270:        pip install .
tests/follow_symlinks.sh:115:pip install .
zstash/parallel.py:132:        # be piped to this queue instead of the screen.
```

setup.py
```diff
@@ -8,10 +8,5 @@
     description="Long term HPSS archiving software for E3SM",
     packages=find_packages(include=["zstash", "zstash.*"]),
     python_requires=">=3.6",
-    install_requires=[
-        "fair-research-login>=0.2.6,<0.3.0",
-        "globus-sdk>=3.0.0,<4.0.0",
-        "six",
-    ],
```
forsyth2, line 15: @tomvothecoder It looks like the tests don't like this part being removed. Am I supposed to add these requirements somewhere else?

xylar (approved):
Yes, at a glance it looks right to me.

setup.py
tomvothecoder: The issue is that the `build` step of the `build_workflow.yml` is using a pip environment and the `install_requires` section from `setup.py`.

You need to replace these lines https://github.com/E3SM-Project/zstash/blob/c0aadc7d7d11b49d3fd60fdc42340b79ee910097/.github/workflows/build_workflow.yml#L38-L41

with the step to cache and create the conda environment ([zppy example](https://github.com/E3SM-Project/zppy/blob/2304054174c723ecd9fb7a94a6a06759d362eee1/.github/workflows/build_workflow.yml#L54-L75)):

```
      - name: Cache Conda
        uses: actions/cache@v3
        env:
          CACHE_NUMBER: 0
        with:
          path: ~/conda_pkgs_dir
          key: ${{ runner.os }}-conda-${{ env.CACHE_NUMBER }}-${{
            hashFiles('conda/dev.yml') }}


      - name: Build Conda Environment
        uses: conda-incubator/setup-miniconda@v2
        with:
          activate-environment: zstash_dev
      miniforge-variant: Mambaforge
          miniforge-version: latest
          use-mamba: true
          mamba-version: "*"
          environment-file: conda/dev.yml
          channel-priority: strict
          auto-update-conda: true
          # IMPORTANT: This needs to be set for caching to work properly!
          use-only-tar-bz2: true
```

forsyth2: @tomvothecoder Thanks! Now I'm getting `sphinx-multiversion 0.2.4  does not exist (perhaps a typo or a missing channel).` on the tests though...

forsyth2: And everything seems to match up with what was done for `zppy`, which did work.

forsyth2: @tomvothecoder Ah, actually `zppy` did not work. I missed a failing GitHub action. Fixing a syntax error in https://github.com/E3SM-Project/zppy/pull/434, I end up with the same error as `zstash` described here.

forsyth2: Yes, my latest commit reverting that change makes the GitHub Actions check pass. I may keep it like that to get this merged and address this issue later in the release testing process.

forsyth2: The checks appeared to pass for https://github.com/E3SM-Project/zppy/pull/434, but after merging the publish-docs Action failed (https://github.com/E3SM-Project/zppy/actions/runs/5284999456/jobs/9563041218: line 1: mamba: command not found). Because of this, I think it's best to hold off on merging this until the `sphinx-multiversion` error is resolved.

xylar:
@forsyth2, I don't have permission to push to this branch.  Could you please cherry-pick the last 2 commits from the following?
https://github.com/xylar/zstash/commits/issue-252

xylar:
> We still want to keep everything functioning if one were to use pip (think users installing a dev version in their custom envs) and this shouldn't be hard at all to obtain while still moving the main actions to mamba. The change in setup.py may break this, but I will need to look more into it.

I think our packages basically don't and never have worked with `pip`.  They require dependencies only available on conda-forge.  I can't say whether that's true for zstash but it's true for everything else (zppy, MPAS-Analysis, mache, E3SM_Diags, etc.)

> I think keeping the docs actions running on pip is okay; I don't see a specific reason here to switch to mamba (but correct me if I am wrong). Using pip will be much lighter on the CI anyway. However, I think Xylar's commits in https://github.com/xylar/zstash/commits/issue-252 will fix the issues and this should run on mamba if we want it to.

I'll leave this up to @tomvothecoder and @forsyth2.  To me, it seems like an undesirable practice to use `pip` for CI and `mamba` for local testing.  I also like to use the automatic API generation features in sphinx and recommend that others use that.  That requires having the package installed before making the docs, so `pip` is out.

setup.py
```diff
@@ -8,10 +8,5 @@
     description="Long term HPSS archiving software for E3SM",
     packages=find_packages(include=["zstash", "zstash.*"]),
     python_requires=">=3.6",
-    install_requires=[
```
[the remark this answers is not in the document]
xylar: We just don't want dependencies to come from `pip`.  I don't remember the exact details.  Hopefully @forsyth2 can point us to the relevant discussion, but these were causing trouble in the recent past.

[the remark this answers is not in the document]
xylar: I think it was just this comment from @tomvothecoder:
https://github.com/E3SM-Project/zppy/issues/397#issuecomment-1512147993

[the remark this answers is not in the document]
forsyth2: > point us to the relevant discussion
I'm not sure if this is what you meant, but these installation requirements were introduced in #154 to enable Globus support. A couple other files subsequently had to be modified, which was done in #186.

forsyth2:
> Could you please cherry-pick the last 2 commits from the following?

Cherry-picked to this branch.

> To me, it seems like an undesirable practice to use `pip` for CI and `mamba` for local testing.

I agree that it probably makes more sense to stick to one tool here.

forsyth2 (commented):
Thanks for the extra commits @xylar! I think this looks good now; should I go ahead and merge?

xylar:
Maybe see how it goes with appy first and merge this only after that's successful...

forsyth2:
Ok, we'll wait to see how https://github.com/E3SM-Project/zppy/pull/438 does

xylar:
@forsyth2, as I just commented here:
https://github.com/E3SM-Project/zppy/pull/438#issuecomment-1595729911
I think we can merge this.  The approach used here worked fine in `zppy`.

PR #275, opened 2023-06-19
add compy (login and compute) and chrlogin to globus endpoints

### LINKED ISSUE #274: Add Compy endpoint to Globus file
Add Compy endpoint to Globus file.

In https://github.com/E3SM-Project/zstash/blob/main/zstash/globus.py, add Compy's Globus endpoint.

mahf708:
fix #274 

- [ ] Especially important to review: a controversial workaround to guide the regex to compy if the the HOSTNAME is set to something starting with compy and the the fqdn is `n*.local`.

### DIFF
--- a/zstash/globus.py
+++ b/zstash/globus.py
@@ -1,6 +1,7 @@
 from __future__ import absolute_import, print_function
 
 import configparser
+import os
 import os.path
 import re
 import socket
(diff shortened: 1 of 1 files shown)


zstash/globus.py
```diff
@@ -22,9 +23,11 @@
 regex_endpoint_map = {
     r"theta.*\.alcf\.anl\.gov": "08925f04-569f-11e7-bef8-22000b9a448b",
     r"blueslogin.*\.lcrc\.anl\.gov": "61f9954c-a4fa-11ea-8f07-0a21f750d19b",
+    r"chrlogin.*\.lcrc\.anl\.gov": "61f9954c-a4fa-11ea-8f07-0a21f750d19b",
     r"b\d+\.lcrc\.anl\.gov": "61f9954c-a4fa-11ea-8f07-0a21f750d19b",
     r"chr.*\.lcrc\.anl\.gov": "61f9954c-a4fa-11ea-8f07-0a21f750d19b",
     r"cori.*\.nersc\.gov": "9d6d99eb-6d04-11e5-ba46-22000b92c6ec",
+    r"compy.*\.pnl\.gov": "68fbd2fa-83d7-11e9-8e63-029d279f7e24",
     r"perlmutter.*\.nersc\.gov": "6bdc7956-fc0f-4ad2-989c-7aa5ee643a79",  # If this doesn't work, use cori
```
forsyth2, line 31: You make a good point, including the compute nodes for Compy. I think Perlmutter compute nodes also don't fit this pattern.

I usually run `zstash` from a login node but I suppose some may find it useful to run a large Globus transfer from a compute node.

forsyth2: @mahf708 The issue I discovered with using compute nodes is that sometimes the Globus transfers take longer than the max wall clock time (12 hours according to https://docs.nersc.gov/jobs/policy/). That is, I think we'd need to set up some sort of restart file system (as E3SM itself does) and use `zstash --update`.

forsyth2:
> * [ ]  Especially important to review: a controversial workaround to guide the regex to compy if the the HOSTNAME is set to something starting with compy and the the fqdn is `n*.local`.

I unfortunately don't know too much about the HOSTNAME issue.

That said, I ran `python -m unittest tests/test_globus.py` on login nodes and compute nodes for Chrysalis, Perlmutter, and Compy (using this branch) and it passes in all 6 cases. (For reference, I ran on compute nodes using `salloc -n 1` on Chrysalis and Compy, and `salloc --constraint=cpu -n 1 --qos=regular` on Perlmutter.)

So, I think this is good to merge. Thanks @mahf708

PR #277, opened 2023-06-21
Fix release workflow not publishing latest tags

tomvothecoder:
I noticed the E3SM Diags release workflow wasn't copying the docs for v2.8.0 over to the gh-pages branch (related [PR](https://github.com/E3SM-Project/e3sm_diags/pull/698)). This caused the docs to break with a 404 if you tried clicking v2.8.0 in the dropdown.

The same thing is happening in zstash with [v1.3.0](https://e3sm-project.github.io/zstash/_build/html/v1.3.0/index.html). I updated the release workflow to correctly copy the tag docs. Once this is merged, we can rerun the release workflow manually to copy v1.3.0 docs over to gh-pages/

**Note, another 1.4.0 release candidate is not required for this fix because this is DevOps related.**

### DIFF
--- a/.github/workflows/release_workflow.yml
+++ b/.github/workflows/release_workflow.yml
@@ -4,6 +4,8 @@ on:
   release:
     types: [published]
 
+  workflow_dispatch:
+
 jobs:
   publish-docs:
     runs-on: ubuntu-latest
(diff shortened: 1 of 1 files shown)


.github/workflows/release_workflow.yml
```diff
@@ -4,6 +4,8 @@ on:
   release:
     types: [published]
 
+  workflow_dispatch:
```
tomvothecoder, line 7: This line allows us to manually run workflows. We need to add this because the release workflow currently only triggers when releasing new versions. However, we want to run it manually to fix the v1.3.0 docs.

PR #309, opened 2023-10-24
Bump to 1.4.1rc1

forsyth2:
Bump to 1.4.1rc1

### DIFF
--- a/conda/meta.yaml
+++ b/conda/meta.yaml
@@ -1,5 +1,5 @@
 {% set name = "zstash" %}
-{% set version = "1.4.0" %}
+{% set version = "1.4.1rc1" %}
 
 package:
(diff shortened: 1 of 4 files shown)


xylar:
@forsyth2, you created a branch with the same name as the eventual tag here, which is a totally logical thing to do.  Unfortunately, it really confuses things once you do create the tag.  I think that's why we're getting a different sha256 hash on conda-forge every time it tries to build from the tag.  Could you delete the branch?  And in the future it is probably best to either put the branch of the same name on your fork instead of the main repo or give it a different name from the eventual tag to avoid trouble.

forsyth2:
Ah, I deleted the branch locally to make the GitHub tag, but I didn't delete the branch on GitHub.

PR #352, opened 2024-09-23
Tweak behaviors of build and release CI workflows

altheaden:
Hello! With this PR I've tweaked some settings that @xylar and I have determined work out better in CI workflows on other repos. I've made the same changes here. These changes include:
1. All references to `mamba`/`mambaforge` have been removed and/or changed to `conda` and `Miniforge3`
2. The timeout for the `pre-commit` job has been upped to 5 minutes

---
Updated template from @forsyth2:

## Issue resolution
- Equivalent to https://github.com/E3SM-Project/e3sm_diags/pull/847, https://github.com/E3SM-Project/zppy/pull/621

This pull request is a minor adjustment that does not affect functional code.

## 1. Does this do what we want it to do?

Objectives:
- Run CI with `Miniforge3` rather than `mamabforge`
- Run CI with a longer timeout for pre-commit checks, to ensure enough running time for these checks to pass.
- [Note: the `cancel_others` change present in the equivalent pull requests is not needed here, since here that parameter isn't in the file to begin with.]

Required:
- [x] Product Management: I have confirmed with the stakeholders that the objectives above are correct and complete.
- [ ] Testing: I have added at least one automated test. Every objective above is represented in at least one test.
  - The existing CI checks on GitHub check this.
- [x] Testing: I have considered likely and/or severe edge cases and have included them in testing.

## 2. Are the implementation details accurate & efficient?

Required:
- [x] Logic: I have visually inspected the entire pull request myself.

## 3. Is this well documented?

Required:
- [ ] Documentation: by looking at the docs, a new user could easily understand the functionality introduced by this pull request.
 - This change does not require new documentation.

## 4. Is this code clean?

Required:
- [x] Readability: The code is as simple as possible and well-commented, such that a new team member could understand what's happening.
- [x] Pre-commit checks: All the pre-commits checks have passed.

PR #351, opened 2024-09-23
Update supported python versions

altheaden:
Support for Python 3.8 (and lower) has been dropped from `conda-forge`, so I've updated the lower bounds of supported versions here accordingly.

---
Updated template from @forsyth2:

## Issue resolution
- Equivalent to https://github.com/E3SM-Project/e3sm_diags/pull/844

This pull request is a minor adjustment that would increment the patch version.

## 1. Does this do what we want it to do?

Objectives:
- Make Python 3.9 the oldest supported Python version in zstash.

Required:
- [x] Product Management: I have confirmed with the stakeholders that the objectives above are correct and complete.
- [ ] Testing: I have added at least one automated test. Every objective above is represented in at least one test.
  - Existing unit tests are sufficient to cover this change.
- [x] Testing: I have considered likely and/or severe edge cases and have included them in testing.

## 2. Are the implementation details accurate & efficient?

Required:
- [x] Logic: I have visually inspected the entire pull request myself.

## 3. Is this well documented?

Required:
- [ ] Documentation: by looking at the docs, a new user could easily understand the functionality introduced by this pull request.
 - This change does not require new documentation.

## 4. Is this code clean?

Required:
- [x] Readability: The code is as simple as possible and well-commented, such that a new team member could understand what's happening.
- [x] Pre-commit checks: All the pre-commits checks have passed.

PR #353, opened 2024-10-02
Update pre-commit dependencies

altheaden:
This PR updates the dependencies within `.pre-commit-config.yaml` to their most recent versions. It also adds pre-commit CI which will hopefully auto-update these dependencies on a monthly basis.

---
Updated template from @forsyth2:

## Issue resolution
- Equivalent to https://github.com/E3SM-Project/zppy/pull/627

This pull request is a minor adjustment that does not affect functional code, aside from minor renamings.

## 1. Does this do what we want it to do?

Objectives:
- Update the dependencies within `.pre-commit-config.yaml` to their most recent versions.
- Include the pre-commit type changes/fixes introduced by this.
- Add pre-commit CI which will hopefully auto-update these dependencies on a monthly basis.

Required:
- [x] Product Management: I have confirmed with the stakeholders that the objectives above are correct and complete.
  - Changes to functional code are minimal. 
- [x] Testing: I have considered likely and/or severe edge cases and have included them in testing.

## 2. Are the implementation details accurate & efficient?

Required:

## 3. Is this well documented?

Required:
  - This change does not require new documentation.

## 4. Is this code clean?

Required:
- [x] Readability: The code is as simple as possible and well-commented, such that a new team member could understand what's happening.

### DIFF
--- a/.pre-commit-config.yaml
+++ b/.pre-commit-config.yaml
@@ -35,7 +35,8 @@ repos:
     rev: v1.11.2
     hooks:
       - id: mypy
-        args: ["--config=setup.cfg"]
+        args: ["--config=setup.cfg", "--install-types", "--non-interactive"]
+        exclude: test*
 
 # https://pre-commit.ci/#configuration
 ci:
--- a/.pre-commit-config.yaml
+++ b/.pre-commit-config.yaml
@@ -4,7 +4,7 @@ fail_fast: true
 
 repos:
   - repo: https://github.com/pre-commit/pre-commit-hooks
-    rev: v3.4.0
+    rev: v5.0.0
     hooks:
       - id: trailing-whitespace
       - id: end-of-file-fixer
@@ -12,19 +12,19 @@ repos:
         exclude: conda/meta.yaml
 
   - repo: https://github.com/psf/black
-    rev: 22.3.0
+    rev: 24.10.0
     hooks:
       - id: black
 
   - repo: https://github.com/PyCQA/isort
-    rev: 5.12.0
+    rev: 5.13.2
     hooks:
       - id: isort
 
   # Need to use flake8 GitHub mirror due to CentOS git issue with GitLab
   # https://github.com/pre-commit/pre-commit/issues/1206
   - repo: https://github.com/pycqa/flake8
-    rev: 3.8.4
+    rev: 7.1.1
     hooks:
       - id: flake8
         args: ["--config=setup.cfg"]
@@ -32,7 +32,12 @@ repos:
         exclude: analysis_data_preprocess
 
   - repo: https://github.com/pre-commit/mirrors-mypy
-    rev: v0.790
+    rev: v1.11.2
     hooks:
       - id: mypy
         args: ["--config=setup.cfg"]
+
+# https://pre-commit.ci/#configuration
+ci:
+  autofix_prs: false
+  autoupdate_schedule: monthly
--- a/zstash/extract.py
+++ b/zstash/extract.py
@@ -254,7 +254,7 @@ def extract_database(
     # This is because we may have different versions of the
     # same file across many tars.
     insert_idx: int
-    iter_inx: int
+    iter_idx: int
     insert_idx, iter_idx = 0, 1
     for iter_idx in range(1, len(matches)):
         # If the filenames are unique, just increment insert_idx.
@@ -350,7 +350,7 @@ def multiprocess_extract(
     workers_to_matches: List[List[FilesRow]] = [[] for _ in range(num_workers)]
     for db_row in matches:
         tar = db_row.tar
-        workers_idx: int
+        worker_idx: int
         for worker_idx in range(len(workers_to_tars)):
             if tar in workers_to_tars[worker_idx]:
                 # This worker gets this db_row.
--- a/zstash/extract.py
+++ b/zstash/extract.py
@@ -350,11 +350,11 @@ def multiprocess_extract(
     workers_to_matches: List[List[FilesRow]] = [[] for _ in range(num_workers)]
     for db_row in matches:
         tar = db_row.tar
-        worker_idx: int
-        for worker_idx in range(len(workers_to_tars)):
-            if tar in workers_to_tars[worker_idx]:
+        workers_idx: int
+        for workers_idx in range(len(workers_to_tars)):
+            if tar in workers_to_tars[workers_idx]:
                 # This worker gets this db_row.
-                workers_to_matches[worker_idx].append(db_row)
+                workers_to_matches[workers_idx].append(db_row)
 
     tar_ordering: List[str] = sorted([tar for tar in tar_to_size])
     monitor: parallel.PrintMonitor = parallel.PrintMonitor(tar_ordering)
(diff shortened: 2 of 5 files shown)


altheaden:
@forsyth2  Following the discussion in https://github.com/E3SM-Project/zppy/pull/627 I started making pre-commit changes here. The `flake8` fixes are straightforward but `mypy` is having issues that I'm not sure how to fix:
```bash
trim trailing whitespace.................................................Passed
fix end of files.........................................................Passed
check yaml...............................................................Passed
black....................................................................Passed
isort....................................................................Passed
flake8...................................................................Passed
mypy.....................................................................Failed
- hook id: mypy
- exit code: 1

zstash/globus.py:13: error: Library stubs not installed for "six.moves.urllib.parse"  [import-untyped]
zstash/hpss.py:7: error: Library stubs not installed for "six.moves.urllib.parse"  [import-untyped]
zstash/hpss.py:7: note: Hint: "python3 -m pip install types-six"
zstash/hpss.py:7: note: (or run "mypy --install-types" to install all missing stub packages)
zstash/hpss.py:7: note: See https://mypy.readthedocs.io/en/stable/running_mypy.html#missing-imports
zstash/hpss_utils.py:180: error: Unused "type: ignore" comment  [unused-ignore]
zstash/hpss_utils.py:219: error: Unused "type: ignore" comment  [unused-ignore]
zstash/create.py:11: error: Library stubs not installed for "six.moves.urllib.parse"  [import-untyped]
zstash/extract.py:353: error: Name "worker_idx" already defined on line 340  [no-redef]
zstash/extract.py:448: error: Unused "type: ignore" comment  [unused-ignore]
zstash/extract.py:543: error: Unused "type: ignore" comment  [unused-ignore]
Found 8 errors in 5 files (checked 29 source files)
```
I believe I've setup the dev environment appropriately (although I could have messed that up), but I am not sure how to approach fixing this. I can keep looking into it but I figured I would update here in case it's just something I'm doing wrong with my setup.

forsyth2:
Hmm I don't think I've seen errors like this before.

.pre-commit-config.yaml
```diff
@@ -4,35 +4,41 @@ fail_fast: true
...
         additional_dependencies: [flake8-isort]
         exclude: analysis_data_preprocess
 
   - repo: https://github.com/pre-commit/mirrors-mypy
-    rev: v0.790
+    rev: v1.11.2
     hooks:
       - id: mypy
-        args: ["--config=setup.cfg"]
+        args: ["--config=setup.cfg", "--install-types", "--non-interactive"]
```
altheaden, line 38: I added these two arguments which auto-fix the importing errors which were apparently added in a previous `mypy` update as discussed here: https://www.github.com/python/mypy/issues/10632#issuecomment-863332698

zstash/extract.py
```diff
@@ -350,7 +350,7 @@ def multiprocess_extract(
     workers_to_matches: List[List[FilesRow]] = [[] for _ in range(num_workers)]
     for db_row in matches:
         tar = db_row.tar
-        workers_idx: int
+        worker_idx: int  # type: ignore
```
altheaden, line 353: This may not be the ideal way to solve this error, but I don't really want to change the code when I'm not familiar with the overall context.

zstash/extract.py
```diff
@@ -350,7 +350,7 @@ def multiprocess_extract(
     workers_to_matches: List[List[FilesRow]] = [[] for _ in range(num_workers)]
     for db_row in matches:
         tar = db_row.tar
-        workers_idx: int
+        worker_idx: int
```
altheaden, line 353: This rename was to match the index variable in the following for loop, but it was obscuring a different error (redundant variable definition) which `mypy` is picking up on now. I'll leave this for somebody who knows the code better to sort out, but it seems like a pretty easy fix.

altheaden:
@forsyth2 This branch is sorted out to the extent that I feel comfortable changing the code, but there is one remaining `mypy` error which I've pointed out above. It looks like a relatively easy fix would be to rename one of the index variables, but I wanted to leave that change for somebody else to make sure it gets fixed correctly. 

I fixed the importing errors `mypy` was complaining about (see my first comment above) but the fix could impact things if it's automatically finding the wrong library stubs. 

Please let me know if there is anything else I can help with, or if I should change anything here.

zstash/extract.py
forsyth2: Ah ok so, the whole relevant block on the `main` branch` is
```
    # Using a greedy approach, populate workers_to_tars.
    for _, tar in enumerate(tar_to_size):
        # The worker with the least work should get the current largest amount of work.
        workers_work: int
        worker_idx: int
        workers_work, worker_idx = heapq.heappop(work_to_workers)
        workers_to_tars[worker_idx].add(tar)
        # Add this worker back to the heap, with the new amount of work.
        worker_tuple: Tuple[float, int] = (workers_work + tar_to_size[tar], worker_idx)
        # FIXME: error: Cannot infer type argument 1 of "heappush"
        heapq.heappush(work_to_workers, worker_tuple)  # type: ignore

    # For worker i, workers_to_matches[i] is a list of
    # matches from the database for it to process.
    workers_to_matches: List[List[FilesRow]] = [[] for _ in range(num_workers)]
    for db_row in matches:
        tar = db_row.tar
        workers_idx: int
        for worker_idx in range(len(workers_to_tars)):
            if tar in workers_to_tars[worker_idx]:
                # This worker gets this db_row.
                workers_to_matches[worker_idx].append(db_row)
```

So yeah once `workers_idx: int` is changed to `worker_idx: int` to actually match the variable in this loop, `mypy` complains because we're attempting to re-defined a variable from the previous loop. It looks like in practice that doesn't actually matter since each occurrence of `worker_idx` stays inside its scoping (i.e. its loop), but `mypy` doesn't have a way to know that.

I think here I'd say to actually keep it defined as `workers_idx: int` and change that one loop accordingly:
```
        workers_idx: int
        for workers_idx in range(len(workers_to_tars)):
            if tar in workers_to_tars[workers_idx]:
                # This worker gets this db_row.
                workers_to_matches[workers_idx].append(db_row)
```

zstash/hpss_utils.py
```diff
@@ -216,7 +216,7 @@ def add_file(
                     blocks += 1
                 # Increase the offset by the amount already saved to the tar
                 # FIXME: error: "TarFile" has no attribute "offset"
-                tar.offset += blocks * tarfile.BLOCKSIZE  # type: ignore
+                tar.offset += blocks * tarfile.BLOCKSIZE
```
forsyth2, line 219: If these lines are passing in `mpy` now, we can probably remove all the error comments on the lines directly above.

altheaden: Sure, I'll take those out right now.

zstash/extract.py
altheaden: OK, I will do that!

altheaden:
@forsyth2 Those changes should be sorted now, looks like everything is passing now as expected.

forsyth2:
@altheaden Ok it looks like I've now merged in all your open PRs (https://github.com/E3SM-Project/zppy/pull/620 / https://github.com/E3SM-Project/zppy/pull/630, https://github.com/E3SM-Project/zppy/pull/621, https://github.com/E3SM-Project/zppy/pull/627 in `zppy` and #351, #352, #353 in `zstash`). Thanks so much for working on these!

PR #384, opened 2025-09-15
Update Python support

forsyth2:
## Summary

Objectives:
- Update Python support to 3.11+

## Small Change


conda/dev.yml
```diff
@@ -6,7 +6,8 @@ dependencies:
   # Base
   # =================
   - pip=22.2.2
-  - python=3.9.13
+  - python=3.11
+  - sqlite
```
forsyth2, line 10: @tomvothecoder How do we determine the constraints (`=`, `>=`, etc.)?

forsyth2 (commented):
@tomvothecoder Can you do a quick review of this please?

<details>

<summary> My testing steps </summary>

```bash
cd ~/ez/zstash
git fetch upstream main
git checkout -b update-python upstream/main
# Make changes to support python3.11+
nersc_conda
rm -rf build
conda clean --all --y
conda env create -f conda/dev.yml -n zstash-py-3-11-20250915
conda activate zstash-py-3-11-20250915
pre-commit run --all-files
python -m pip install .
# FAILED (errors=4, skipped=1)
# Globus test was probably skipped because authentications were missing.
# ERROR: testCheckParallelVerboseMismatch (tests.test_check_parallel.TestCheckParallel.testCheckParallelVerboseMismatch)
# ERROR: testCheckParallelVerboseMismatchHPSS (tests.test_check_parallel.TestCheckParallel.testCheckParallelVerboseMismatchHPSS)
# ERROR: testCheckVerboseMismatch (tests.test_check.TestCheck.testCheckVerboseMismatch)
# ERROR: testCheckVerboseMismatchHPSS (tests.test_check.TestCheck.testCheckVerboseMismatchHPSS)
# FileNotFoundError: [Errno 2] No such file or directory: 'sqlite3'
python -m unittest tests.test_check.TestCheck.testCheckVerboseMismatch
# FileNotFoundError: [Errno 2] No such file or directory: 'sqlite3'
git add -A
git commit -m "Update Python support"

git checkout -b test-main upstream/main
rm -rf build
conda clean --all --y
conda env create -f conda/dev.yml -n zstash-py-3-9-20250915
conda activate zstash-py-3-9-20250915
pre-commit run --all-files
python -m pip install .
python -m unittest tests.test_check.TestCheck.testCheckVerboseMismatch # OK
python -m unittest tests.test_check.TestCheck.testCheckVerboseMismatch # OK
python -m unittest tests.test_check_parallel.TestCheckParallel.testCheckParallelVerboseMismatchHPSS # OK
python -m unittest tests.test_check_parallel.TestCheckParallel.testCheckParallelVerboseMismatch # OK

git checkout update-python
git push upstream update-python
# https://github.com/E3SM-Project/zstash/pull/384/files

# Add `- sqlite`
rm -rf build
conda clean --all --y
conda env create -f conda/dev.yml -n zstash-py-3-11-sqlite-20250915
conda activate zstash-py-3-11-sqlite-20250915
pre-commit run --all-files
python -m pip install .
python -m unittest tests.test_check.TestCheck.testCheckVerboseMismatch # OK
python -m unittest tests.test_check.TestCheck.testCheckVerboseMismatchHPSS # OK
python -m unittest tests.test_check_parallel.TestCheckParallel.testCheckParallelVerboseMismatchHPSS # OK
python -m unittest tests.test_check_parallel.TestCheckParallel.testCheckParallelVerboseMismatch # OK
git add -A
git commit -m "Add explicit sqlite dependency"
```

</details>

conda/dev.yml
```diff
@@ -6,7 +6,8 @@ dependencies:
   # Base
   # =================
   - pip=22.2.2
-  - python=3.9.13
+  - python=3.11
```
tomvothecoder, line 9: The `dev.yml` constraints should align to the conda-forge [recipe](https://github.com/conda-forge/zstash-feedstock/blob/main/recipe/meta.yaml). Currently, the constraints are too restrictive with exact versions pinned.

Follow this for more guidance if needed: https://github.com/E3SM-Project/e3sm_diags/pull/1006

forsyth2:
Ok, appear to have everything working now: 

```bash
python -m ensurepip --upgrade
python -m pip install --upgrade setuptools wheel
git fetch upstream
# Get rid of fair-research-login dependency by using latest `main`:
git rebase upstream/main
rm -rf build
conda clean --all --y
conda env create -f conda/dev.yml -n zstash-20250916v2
conda activate zstash-20250916v2
pre-commit run --all-files
git add -A
python -m pip install .
# Still getting:
# AttributeError: module 'pkgutil' has no attribute 'ImpImporter'. Did you mean: 'zipimporter'?

# Use conda's python to ensure pip in the right environment
python -m ensurepip --upgrade
python -m pip install --upgrade --force-reinstall setuptools wheel pip
python -m pip install .
# Success
# Success
```

forsyth2:
While I got it working for me on Perlmutter, the GitHub Actions were still failing. After substantial iteration with Claude arrived at some fixes. This now runs GitHub Actions on Python 3.11 and 3.12.

<details>

<summary> Steps </summary>

```bash
# Add Claude's Matrix testing suggestion
pre-commit run --all-files
git add -A
python -m pip install .
# Success
# Success
git commit -m "Add matrix testing"
git push upstream update-python
# Still fails on GitHub Actions

# Additional fixes to the github workflow
pre-commit run --all-files
git add -A
git commit -m "Additional fixes"
git push upstream update-python
# Still fails on GitHub Actions

# More changes
pre-commit run --all-files
git add -A
git commit -m "Fixes 3rd try"
git push upstream update-python
# 3.11 alone works!

# Add back matrix testing -- 3.11, 3.12, 3.13
pre-commit run --all-files
git add -A
python -m pip install .
# Success
# Success
git commit -m "Fixes 4th try"
git push upstream update-python
# 3.11, 3.12 get cancelled. 3.13 errors.

# Remove 3.13
pre-commit run --all-files
git add -A
git commit -m "Fixes 5th try"
git push upstream update-python
# 3.11, 3.12 work!
```

</details>

conda/dev.yml
```diff
@@ -1,12 +1,13 @@
 name: zstash_dev
 channels:
   - conda-forge
-  - defaults
+  - nodefaults  # This replaces 'defaults' to avoid Anaconda commercial repos
```
forsyth2, line 4: In one of the fix attempts, GitHub Actions failed because of some licensing issue, so Claude suggested this fix. Although, looking at the final PR diff, it seems strange `defaults` worked fine before...

forsyth2 (commented):
@tomvothecoder Can you please review the latest version of changes? Thanks! 

I addressed your comments and made some other substantial changes to get `pip install` working & GitHub Actions passing.

If this looks good, I can make similar changes for the equivalent PRs at https://github.com/E3SM-Project/zppy/pull/735 & https://github.com/E3SM-Project/zppy-interfaces/pull/33.

forsyth2:
Re: Python 3.13 support -- `python =3.13 * is not installable because it conflicts with any installable versions previously reported.` appears in https://github.com/E3SM-Project/zstash/actions/runs/17783224150/job/50546110928. The newly added matrix testing only tests for 3.11 and 3.12.

I still need to add matrix testing to zppy/zppy-interfaces. Previously, we've just tested on one version.

conda/dev.yml
tomvothecoder: This is based on the conda-forge recipe and compatibility with other packages. If things break with new package versions, that suggests updating the constraints accordingly.

setup.cfg
```diff
@@ -46,7 +46,7 @@ exclude =
     venv
 
 [mypy]
-python_version = 3.9
+python_version = 3.13
```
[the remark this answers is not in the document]
tomvothecoder: Latest stable version of Python.

conda/dev.yml
tomvothecoder: I just specify `conda-forge` and remove `defaults`. I don't know whether `nodefaults` is needed or not.

.github/workflows/build_workflow.yml
```diff
@@ -19,21 +19,24 @@ jobs:
...
       # If you commit with the `--no-verify` flag, this check may fail.
       - name: Install and Run Pre-commit
-        uses: pre-commit/action@v3.0.0
+        uses: pre-commit/action@v3.0.1
 
   build:
     runs-on: ubuntu-latest
+    strategy:
+      matrix:
+        python-version: [3.11, 3.12]
```
tomvothecoder, line 39:
```suggestion
        python-version: [3.11, 3.12, 3.13]
```

.github/workflows/build_workflow.yml
```diff
@@ -19,21 +19,24 @@ jobs:
       - name: Checkout Code Repository
         uses: actions/checkout@v3
 
-      - name: Set up Python 3.9
+      - name: Set up Python 3.11
```
tomvothecoder, line 22:
```suggestion
      - name: Set up Python 3.13
```

.github/workflows/build_workflow.yml
```diff
@@ -19,21 +19,24 @@ jobs:
...
         uses: actions/setup-python@v4
         with:
-          python-version: 3.9
+          python-version: 3.11
```
tomvothecoder, line 25:
```suggestion
          python-version: 3.13
```

tomvothecoder (commented):
My quick review. TLDR: Update dev.yml and try with Python 3.13 again.

forsyth2 (commented):
@tomvothecoder, thanks for your earlier tips. Can you please review this latest iteration? It appears to be passing all the GitHub Actions (for Python 3.11, 3.12, 3.13) now. Thanks!

.github/workflows/build_workflow.yml
```diff
@@ -100,8 +117,9 @@ jobs:
           miniforge-variant: Miniforge3
           miniforge-version: latest
           environment-file: conda/dev.yml
-          channel-priority: strict
+          channel-priority: flexible  # Changed from strict to flexible
           auto-update-conda: true
+          python-version: "3.11"  # Use stable Python version for docs
```
tomvothecoder, line 122:
```suggestion
          python-version: "3.13"  # Use stable Python version for docs
```

zstash/hpss_utils.py
```diff
@@ -281,43 +282,67 @@ def add_file(
...
+        sys.version_info >= (3, 13)
+        and (tarinfo.isfile() or tarinfo.islnk())
+        and tarinfo.size > 0
+    ):
+        # Python 3.13+ requires fileobj for non-empty regular files
+        with open(file_name, "rb") as fileobj:
+            tar.addfile(tarinfo, fileobj)
+    else:
+        # Original code - unchanged
+        tar.addfile(tarinfo)
```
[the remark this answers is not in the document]
forsyth2: @andrewdnolan ^ This is what I mentioned in the EZ meeting re: Python 3.13 support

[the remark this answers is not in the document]
tomvothecoder: > E.g., is it possible that tars `create`d under 3.12 might not be the same size as those `extract`ed under 3.13? I'm not sure.

Unit tests and ad-hoc tests on real-world data would help verify this here. If the tests pass between versions, even with logical changes, then you're good to go.

Otherwise a quick sanity test is to write a Python script using both tar functions and comparing the results.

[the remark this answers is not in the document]
xylar: @forsyth2, we need to support the range of python version stated here:
https://e3sm.atlassian.net/wiki/spaces/DOC/pages/129732419/Packages+in+the+E3SM+Unified+conda+environment#e3sm-unified-1.12.0
```
python >=3.11,<3.14
```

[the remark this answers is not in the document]
xylar: Similarly, presumably different logic should not be needed below for python >=3.13 and python <3.13.

xylar (approved):
@forsyth2, the `tar` related code changes look good to me.  I think not having separate blocks for different python versions is a good direction to take.

I'm approving the PR on the assumption that you and @tomvothecoder will work out how you want to handle the pre-commit dependencies both in the pre-commit config file and in the conda environment.

conda/dev.yml
```diff
@@ -1,31 +1,30 @@
...
-  - black=24.10.0
-  - flake8=7.1.1
-  - flake8-isort=6.1.1
-  - mypy=1.11.2
-  - pre-commit=4.0.1
-  - tbump=6.9.0
+  - black >=24.0.0
+  - flake8 >=7.0.0
+  - flake8-isort >=6.0.0
+  - mypy >=1.11.0
```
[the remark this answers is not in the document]
tomvothecoder: > @tomvothecoder, is it your philosophy that these need to match `.pre-commit-config.yaml` exactly?

Correct, these should align exactly with `.pre-commit-config.yaml` with exact version pinned as you mentioned before.

zstash/hpss_utils.py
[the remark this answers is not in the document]
tomvothecoder: > It's very hard to do true "unit" testing on `zstash` because almost everything requires I/O (i.e., you can't test abstract functions with no side effects).

I'm not familiar with the functionalities of `zstash` exactly so I might be off here, but I'd imagine you can write unit tests that utilize `pytest`s [temporary directory](https://docs.pytest.org/en/6.2.x/tmpdir.html) to write/read files which you then pass to `zstash` functions to get a result. Of course this doesn't consider different machines filesystems and what not, but it at least tests the direct functions to ensure it doesn't unexpectedly break between changes such as swapping `tar` functions.

Just something to think about for future test suite enhancements.

conda/meta.yaml
```diff
@@ -16,14 +16,14 @@ build:
...
 
   run:
-    - python >=3.9
-    - fair-research-login >=0.2.6,<0.3.0
-    - globus-sdk >=3.0.0,<4.0.0
-    - six
+    - python >=3.11,<3.14
+    - globus-sdk=3.15.0
+    - six=1.16.0
+    - sqlite
```
tomvothecoder, line 26: @forsyth2 Did you mean to constrain exact versions of `globus-sdk` and `six` here?

On a side-note, I don't think this file is used anymore? I'm pretty sure this file was used to host `zstash` on the `e3sm` channel, which we no longer do since adopting `conda-forge`.

I think you can just delete this file  (`meta.yaml`) if that's the case.

tomvothecoder (changes requested):
My review, remaining fixes:

1. Align QA tooling versions, specifically conda environments with `pre-commit` using my key points at the bottom of this review.
2. Finalize the `tar` related code changes.

After that, it should be good to go.

---
> @forsyth2, the tar related code changes look good to me. I think not having separate blocks for different python versions is a good direction to take.

I agree with @xylar here.

> I'm approving the PR on the assumption that you and @tomvothecoder will work out how you want to handle the pre-commit dependencies both in the pre-commit config file and in the conda environment.

@forsyth2 The key points are:

1. **Pin exact versions** of QA tools
   * Don’t use version ranges—always specify the exact version.
   * This guarantees consistent behavior between your local development environment and CI.
2. **Use the same versions across all environments**
   * QA tools should match exactly between conda environments (from **conda-forge**) and pre-commit hooks (from **PyPI**).
   * If `pre-commit autoupdate` pulls a version that’s newer than what’s available on conda-forge, you should fall back to the **latest version on conda-forge** instead for both `pre-commit` and Conda environments.

zstash/hpss_utils.py
```diff
@@ -281,43 +280,29 @@ def add_file(
...
-                blocks: int
-                remainder: int
-                blocks, remainder = divmod(tarinfo.size, tarfile.BLOCKSIZE)
-                if remainder > 0:
-                    null_bytes: bytes = tarfile.NUL
-                    # Write null_bytes to get the last block to tarfile.BLOCKSIZE
-                    fileobj.write(null_bytes * (tarfile.BLOCKSIZE - remainder))
-                    blocks += 1
-                # Increase the offset by the amount already saved to the tar
-                tar.offset += blocks * tarfile.BLOCKSIZE
```
forsyth2, line 317: It appears we don't actually need to be tracking/updating `tar.offset` here.

zstash/hpss_utils.py
```diff
@@ -281,43 +280,29 @@ def add_file(
...
     if tarinfo.islnk():
         tarinfo.size = os.path.getsize(file_name)
+
     # Add the file to the tar
-    tar.addfile(tarinfo)
+    if (tarinfo.isfile() or tarinfo.islnk()) and tarinfo.size > 0:
+        with open(file_name, "rb") as fileobj:
+            tar.addfile(tarinfo, fileobj)
+    else:
+        tar.addfile(tarinfo)
```
forsyth2, line 289: In the newly supported versions of Python, `addfile` writes to `fileobj`.

forsyth2 (commented):
@golaz It looks like my original comment got removed because the code changed. Can you please review the diff for `zstash/hpss_utils.py`? Thanks!

zstash/hpss_utils.py
xylar: My reading of the documentation is that yes, the tar file is updated already with the `tar.add_file()` call when a `fileobj` is passed:
https://docs.python.org/3/library/tarfile.html

The advantage of the original implementation was probably that each file got opened once only and was added to the archive at the same time as computing its md5 hash.  The new code reopens the file to create the md5 hash but I think that's fine (and seemingly unavoidable with the new API).

golaz (changes requested):
Cannot approve. I don't understand why we would remove a key functionality of zstash with likely severe performance impact simply to update to a new version of Python.

zstash/hpss_utils.py
forsyth2: Claude's feedback:
========

If the old implementation breaks on Python 3.13 (likely due to changes in how `tarfile` handles internal file objects), then you're facing a **compatibility vs. performance trade-off**.

## Options to Consider

### 1. **Accept the performance hit** (pragmatic choice)
- The new implementation works on Python 3.13
- Performance regression may be acceptable depending on:
  - Typical file sizes in your use case
  - Available I/O bandwidth (SSDs make double-reads less painful)
  - Whether this is a bottleneck in practice

### 2. **Implement a hybrid solution** (best of both worlds)
```python
def add_file(tar, file_name, follow_symlinks):
    offset = tar.offset
    tarinfo = tar.gettarinfo(file_name)
    
    if tarinfo.islnk():
        tarinfo.size = os.path.getsize(file_name)
    
    md5 = None
    
    # For files/hardlinks, stream data while computing hash
    if (tarinfo.isfile() or tarinfo.islnk()) and tarinfo.size > 0:
        hash_md5 = hashlib.md5()
        
        # Create a wrapper that computes hash while data passes through
        class HashingFileWrapper:
            def __init__(self, fileobj, hasher):
                self.fileobj = fileobj
                self.hasher = hasher
            
            def read(self, size=-1):
                data = self.fileobj.read(size)
                if data:
                    self.hasher.update(data)
                return data
        
        with open(file_name, "rb") as f:
            wrapper = HashingFileWrapper(f, hash_md5)
            tar.addfile(tarinfo, wrapper)
        
        md5 = hash_md5.hexdigest()
    else:
        tar.addfile(tarinfo)
    
    size = tarinfo.size
    mtime = datetime.utcfromtimestamp(tarinfo.mtime)
    return offset, size, mtime, md5
```

This wrapper approach:
- ✅ Uses the proper Python 3.13-compatible API
- ✅ Maintains single-pass streaming
- ✅ Computes hash during the tar write operation
- ✅ Clean, maintainable code

### 3. **Version-specific implementations**
Use the old code for Python <3.13 and new code for 3.13+, but this adds maintenance burden.

## My Recommendation

**Try the hybrid solution first** (Option 2). It should give you Python 3.13 compatibility while preserving the streaming performance that Reviewer 2 correctly identified as important. If `tar.addfile()` properly uses the file-like object's `read()` method, this should work perfectly.

If the wrapper approach doesn't work for some reason, then you'll need to accept the performance regression as the cost of Python 3.13 support, but at least you'll have tried to preserve the optimization.

forsyth2: @xylar @golaz Would that option 2 ("hybrid solution") work? If not, I'm not sure what else we could do.

xylar: @forsyth2, I could get behind the concept of option 2.  I would prefer not to have a class embedded in a function (ever) .

We would also need to make sure that all `tar.addfile()` requires of its `fileobj` parameter is a class that implements the `read()` method.  A little searching suggests that may be the case.

chengzhuzhang: Sorry I haven’t followed this thread too closely. Which Python version is causing the incompatibility with zstash? I think it’s a solid achievement that we could update from 3.10 (now end-of-life) to 3.11 (or higher, if our tools are ready). In this case, it seems zstash isn’t ready yet, and we are time-constrained to update it with thorough testing.

Can we target Python 3.11 or 3.12 (depending on zstash readiness) for this upcoming unified? Both are still widely used, and 3.11 will continue receiving security updates until October 2027. Again, for a tool like E3SM-unified, I believe we should continue prioritizing maximum compatibility.

forsyth2: >  I could get behind the concept of option 2.
> Hopefully, we can get things there for Unified, since testing should start on Monday. 

I've implemented this in the latest commit: https://github.com/E3SM-Project/zstash/pull/384/commits/cc3723726454a14ffe9831ac4f270a34083951f5. It required an additional change to handle empty files. It now passes the GitHub Actions for 3.11, 3.12, and 3.13.

I believe this means this PR should be good to merge. @xylar @golaz Can you please re-confirm based on this latest commit?

> zstash has not been getting the TLC it has needed over the years to remain compatible with modern python
> Hopefully, we can also find a longer term solution for zstash to have a "champion".

Yes, I agree with all this. After the Unified testing period, perhaps it would be useful to have a team discussion about this. The problem is certainly not a lack of ideas re: code maintainability (e.g., removing all those global variables or improving the testing).

> Which Python version is causing the incompatibility with zstash?

3.13

> find out if zstash works with python 3.12 and without requiring these changes.

It does. I only ran into errors when I started matrix-testing on 3.13.

> All our software must, must, must support the 3 latest stable versions of python at all times.

This is good to keep in mind, thanks.

forsyth2:
Thanks @xylar, see https://github.com/E3SM-Project/zstash/pull/388#pullrequestreview-3301069544

PR #393, opened 2025-10-17
Improve tests

### LINKED ISSUE #385: [Feature]: Improve zstash testing
### How will this affect the next version number?

New feature (increment MINOR version)

### Is your feature request related to a problem?

Current testing has a number of problems:
- So-called "unit" testing is really more integration testing because it involves I/O rather than pure function testing. This leads to two points:
  - It may perhaps be better to convert the "unit" tests to bash scripts that check output. For example, one implementation of this is `tests/scripts/globus_auth.bash`.
  - Actual unit tests should test functions used by `zstash` rather than the actual `zstash` commands called from the command line.
- The current test framework uses `unittest` rather than `pytest`. I recall trying to modernize that in the past, but running into problems.
- The current test framework always uses the GitHubActions or `conda/dev.yml` Python version. We should simulate running `zstash create` with an older version of Python and/or `zstash` and then running `zstash extract` on a later version. For example, see the thread at https://github.com/E3SM-Project/zstash/pull/384#discussion_r2389348376. It's a common use case that someone would be extracting data that was archived a while ago.

### Describe the solution you'd like

Described above.

### Describe alternatives you've considered

_No response_

### Additional context

_No response_

forsyth2:
## Summary

This PR better structures existing tests, so it is clearer how to run each of them. It is _not_ significantly changing _what_ we're testing.

Objectives:
- Restructure testing directories to be clearer to follow.
- Add test assertions to test files without them (i.e., make it so output doesn't have to be manually reviewed).
- Make it so the tests files don't have to be modified to run (e.g., by changing `try_#` paths)
- Convert some tests to `pytest` (The tests in `tests/integration/python_tests` will require more effort to convert over).
- Change `conda` setup accordingly to get tests working.

Issue resolution:
- Closes #385

## Small Change

### DIFF
--- a/tests/README.md
+++ b/tests/README.md
@@ -100,7 +100,11 @@ time ./symlinks.sh # NOTE: you will have to change out paths for your username
 # real	0m1.346s
 # No errors
 
-# TODO: Handle blocking_test_scripts
+cd blocking_test_scripts
+# Review README_TEST_BLOCKING
+# This uses "12 piControl ocean monthly files, 49 GB",
+# so processing may take a long time.
+# TODO (later PR): Confirm this test works
 ```
 
 ## Testing with GitHub Actions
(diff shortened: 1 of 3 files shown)


forsyth2 (commented):
This PR is almost exclusively test changes. Since we are in the E3SM Unified testing period, and iteratively updating tests accordingly, I'm just going to go ahead and merge this.

The only non-testing attribute changed is that the permissible versions of `globus-sdk` needed an upper bound to pass tests. As such, I will make a `zstash rc3` after merging this PR.

PR #402, opened 2025-11-18
Add Python 3.14 support

forsyth2:
## Summary

Important: this should be merged to `main` _after_ the production release of `zstash v1.5.0`.

Objectives:
- Add support for Python 3.14

Issue resolution:
- Equivalent of https://github.com/E3SM-Project/zppy/pull/757, https://github.com/E3SM-Project/zppy-interfaces/pull/44

## Small Change

### DIFF
--- a/zstash/extract.py
+++ b/zstash/extract.py
@@ -7,6 +7,7 @@ import heapq
 import logging
 import multiprocessing
 import os.path
+import queue
 import re
 import sqlite3
 import sys
@@ -389,12 +390,21 @@ def multiprocess_extract(
     # because we'll be in this loop until completion.
     failures: List[FilesRow] = []
     while any(p.is_alive() for p in processes):
-        while not failure_queue.empty():
-            failures.append(failure_queue.get())
+        try:
+            while True:
+                failures.append(failure_queue.get_nowait())
+        except queue.Empty:
+            pass
         time.sleep(0.01)
 
-    while not failure_queue.empty():
-        failures.append(failure_queue.get())
+    # Drain any remaining failures after all processes have exited.
+    try:
+        while True:
+            failures.append(failure_queue.get_nowait())
+    except queue.Empty:
+        pass
+
+    manager.shutdown()
 
     # Sort the failures, since they can come in at any order.
     failures.sort(key=lambda t: (t.name, t.tar, t.offset))
@@ -547,18 +557,16 @@ def extractFiles(  # noqa: C901
             # let the process know.
             # This is to synchronize the print statements.
 
-            # Wait for turn before processing this tar
             if multiprocess_worker:
-                multiprocess_worker.print_monitor.wait_turn(
-                    multiprocess_worker, files_row.tar
-                )
                 multiprocess_worker.set_curr_tar(files_row.tar)
 
-            # Use args.hpss directly
+            # Use args.hpss, falling back to config.hpss when not provided
             if args.hpss is not None:
                 hpss: str = args.hpss
+            elif config.hpss is not None:
+                hpss = config.hpss
             else:
-                raise TypeError("Invalid args.hpss={}".format(args.hpss))
+                raise TypeError("Invalid config.hpss={}".format(config.hpss))
 
             tries: int = args.retries + 1
             # Set to True to test the `--retries` option with a forced failure.
@@ -735,9 +743,9 @@ def extractFiles(  # noqa: C901
         cur.close()
         con.close()
 
-        # Add the failures to the queue.
-        # When running with multiprocessing, the function multiprocess_extract()
-        # that calls this extractFiles() function will return the failures as a list.
+    # Add the failures to the queue.
+    # When running with multiprocessing, the function multiprocess_extract()
+    # that calls this extractFiles() function will return the failures as a list.
     if multiprocess_worker:
         for f in failures:
             multiprocess_worker.failure_queue.put(f)
--- a/zstash/parallel.py
+++ b/zstash/parallel.py
@@ -38,6 +38,11 @@ class PrintMonitor(object):
         # Store the ordered list of tars
         self._tars_list: List[str] = tars_to_print
 
+        # Precomputed mapping from tar name to its position in the ordered list.
+        self._tar_to_index: Dict[str, int] = {
+            tar: i for i, tar in enumerate(tars_to_print)
+        }
+
         # Use a simple counter to track which tar we're on
         self._current_tar_index: multiprocessing.managers.ValueProxy = manager.Value(
             "i", 0
@@ -52,10 +57,9 @@ class PrintMonitor(object):
         """
         Wait until it's this worker's turn to process workers_curr_tar.
         """
-        try:
-            tar_index = self._tars_list.index(workers_curr_tar)
-        except ValueError:
-            return
+        if workers_curr_tar not in self._tar_to_index:
+            raise RuntimeError("Tar {} not in ordered list".format(workers_curr_tar))
+        tar_index = self._tar_to_index[workers_curr_tar]
 
         attempted = False
         while True:
@@ -75,10 +79,9 @@ class PrintMonitor(object):
         A worker has finished printing output for workers_curr_tar.
         Advance to the next tar in the sequence.
         """
-        try:
-            tar_index = self._tars_list.index(workers_curr_tar)
-        except ValueError:
-            return
+        if workers_curr_tar not in self._tar_to_index:
+            raise RuntimeError("Tar {} not in ordered list".format(workers_curr_tar))
+        tar_index = self._tar_to_index[workers_curr_tar]
 
         with self._lock:
             if self._current_tar_index.value == tar_index:


forsyth2:
After some testing the branch at various stages, Claude was able to generate this summary table: 

---

<html><head></head><body>

Case | Test Result | Comments on Output (Ordering)
-- | -- | --
Case 1: Base (0 commits) | ✅ PASS | Tar files extracted sequentially in correct order: 000000.tar → 000001.tar → 000002.tar → 000003.tar → 000004.tar. No ordering issues detected.
Case 2: +1 commit (Python 3.14 support) | ❌ FAIL | Fatal error during parallel extract: TypeError: cannot pickle 'sqlite3.Cursor' object. No extraction output produced. Test fails because expected file extraction messages are absent. The sqlite3 cursor cannot be serialized for multiprocessing.
Case 3: +7 commits (functionality fixes) | ❌ FAIL | Ordering violation due to "Worker waiting" debug messages. The tar order array is [0, 0, 1, 1, 2, 2, 3, 3, 4, 4, 0, 0, 0, 0, 0, 1, 2, 3, ...] instead of [0, 0, 0, 0, 0, 0, 0, 1, 1, 1, ...]. The first part [0, 0, 1, 1, 2, 2, 3, 3, 4, 4] comes from the initial non-verbose extraction output. The problem is the debug section contains DEBUG: Worker waiting for tar 000001.tar (index 1), current index is 0 which adds extra tar number mentions before tar 0 completes in the parsed output, breaking sequential ordering.
Case 4: +4 commits (major refactor) | ✅ PASS | Tar files extracted in correct sequential order with proper synchronization. All expected "Extracting" messages present. Test passes completely - both functional correctness and ordering requirements met.
Case 5: -4 commits, +1 commit (comment out "Worker waiting" line) | ❌ FAIL | Ordering still violated, but differently. Tar order is now [0, 0, 1, 1, 2, 2, 3, 3, 4, 4, 0, 0, 0, 0, 1, 1, 1, 1, 2, 2, 2, 2, ...] instead of expected [0, 0, 0, 0, 0, 0, 1, 1, 1, 1, 1, 1, ...]. Removing "Worker waiting" messages eliminated the early interleaving (no more 0, 0, 1, 1, 2, 2, 3, 3, 4, 4, 0, 0, 0, 0, 0, 1, 2, 3), but the fundamental issue remains: the first extraction block [0, 0, 1, 1, 2, 2, 3, 3, 4, 4] is still counted, followed by the debug write() calls [0, 0, 0, 0, 1, 1, 1, 1, ...], creating duplicate/interleaved counting.
Case 6: -1 commit, +1 commit (remove both "Worker waiting" and "write()" debug lines) | ❌ FAIL | Cleaner ordering violation. Tar order is [0, 0, 1, 1, 2, 2, 3, 3, 4, 4, 0, 0, 1, 1, 2, 2, 3, 3, 4, 4] instead of [0, 0, 0, 0, 1, 1, 1, 1, 2, 2, 2, 2, 3, 3, 3, 3, 4, 4, 4, 4]. Both debug message types removed, resulting in simpler pattern: first extraction produces [0, 0, 1, 1, 2, 2, 3, 3, 4, 4] (2 tar mentions per archive), second extraction also produces [0, 0, 1, 1, 2, 2, 3, 3, 4, 4] (2 tar mentions per archive). Debug section now only shows: DEBUG: Worker got turn for tar 000000.tar! and DEBUG: Set curr_tar to 000000.tar (repeated for each tar). The root cause is clear: two separate extraction runs are being counted, not interleaved debug output.


<p><strong>Key insight from Case 6</strong>: Removing all the verbose debug logging reveals the actual problem—the test is capturing output from <strong>two consecutive extraction runs</strong> (both the first <code>+ zstash extract -v --hpss=none --workers=3</code> and what appears to be a second run), and the regex is counting tar mentions from both, creating the doubled pattern.</p></body></html>

---

**Yes, the major refactor in Case 4 was necessary.**

Here's why:

**The Root Problem (revealed by Case 6):**
The test captures output from what appears to be **two extraction runs**, and `compare_tar_orders()` counts tar mentions from both runs. Each run mentions each tar file twice (e.g., `000000.tar` appears in "exists. Checking expected size" and "Opening tar archive"), resulting in the pattern `[0, 0, 1, 1, 2, 2, 3, 3, 4, 4, 0, 0, 1, 1, 2, 2, 3, 3, 4, 4]`.

**Why Cases 3, 5, and 6 all fail:**
- **Case 3**: Debug messages added MORE tar mentions to an already-broken output capture
- **Case 5**: Removed "Worker waiting" messages but still had "write()" debug messages AND the double-extraction problem
- **Case 6**: Removed all verbose debug but still had the double-extraction problem

**Why Case 4 succeeds:**
The major refactor must have fundamentally changed **how output is captured/produced** during parallel extraction. It likely:
1. Ensured only ONE extraction run's output is captured by the test, OR
2. Changed the output mechanism so that tar file mentions appear in strict sequential order even across multiple processing phases, OR
3. Modified the test infrastructure to properly isolate/reset between runs

Simply removing debug messages (Cases 5 & 6) only treated symptoms. The refactor in Case 4 addressed the architectural issue of how parallel workers coordinate their console output and/or how the test captures that output.

**Conclusion**: Yes, the major refactor was necessary because the problem wasn't just verbose logging—it was a fundamental issue with output ordering/capture in the parallel execution model.

--- 

Looking at the test file, I need to **revise my previous answer**. The test structure shows there are actually **TWO separate `zstash extract` commands run sequentially**:

1. **First run** (line 65-78): `zstash extract -v --hpss={} --workers=3` (with verbose flag)
2. **Second run** (line 80-93): `zstash extract --hpss={} --workers=3` (without verbose flag)

Both runs call `self.compare_tar_orders(output + err)` to check ordering.

**Revised Understanding:**

**The Root Problem:**
Each individual extraction run is producing output where tar files are mentioned out of order. The pattern `[0, 0, 1, 1, 2, 2, 3, 3, 4, 4]` from a single run shows that:
- Tar 0 is mentioned twice (e.g., "000000.tar exists" + "Opening tar archive 000000.tar")
- Tar 1 is mentioned twice
- Etc.

But they appear interleaved: tar 0 (twice), then tar 1 (twice), then tar 2 (twice)... instead of all tar 0 mentions grouped together.

**Why Cases 3, 5, and 6 fail:**
Looking at Case 6's output more carefully, the pattern `[0, 0, 1, 1, 2, 2, 3, 3, 4, 4, 0, 0, 1, 1, 2, 2, 3, 3, 4, 4]` is:
- **First half**: `[0, 0, 1, 1, 2, 2, 3, 3, 4, 4]` from the verbose run
- **Second half**: `[0, 0, 1, 1, 2, 2, 3, 3, 4, 4]` from the non-verbose run

Each run individually has the wrong ordering! The workers are processing tars concurrently but their output isn't properly serialized.

**Why Case 4 succeeds:**
The major refactor **was absolutely necessary** because it fixed the fundamental parallel output coordination problem. It likely implemented proper output serialization so that within each extraction run, all mentions of tar N complete before any mention of tar N+1 begins, giving the correct pattern `[0, 0, 0, 0, 1, 1, 1, 1, ...]` for each run.

**Conclusion**: Yes, the major refactor was necessary. The problem wasn't about capturing two runs—that's intentional test design. The problem was that **each individual parallel extraction run** wasn't maintaining sequential tar ordering in its output, and the refactor fixed that core parallelization/synchronization issue.

forsyth2:
It turns out that case 3 was in fact simply failing because of the added debug lines.

Two important lessons:
- Claude appears to be over-eager to engineer overly-complex solutions. Here, we see that the major refactor (case 4) turned out to be completely unnecessary. It is evidently extremely important that in AI prompts we 1) specify to change as little code as possible and 2) include any key pieces of code that are likely to be the issue (so it can focus in on that).
  - Jill has noted previously [in another PR](https://github.com/E3SM-Project/zstash/pull/411#issuecomment-3638640732) how Claude had over-engineered a solution.
  - (While Claude did note "This wasn't over-engineering" [above](https://github.com/E3SM-Project/zstash/pull/402#issuecomment-3812929929), that was because I had said in the prompt the refactor had turned out necessary. That is, it wasn't the same context window where Claude had over-engineered the original solution).
- The zstash output review tests are extremely sensitive to printed output. Simply introducing new debug lines can easily break them.
  - I can't find the discussion to link, but Tony had previously run into this when introducing debug lines.

forsyth2:
I've updated the commits. I removed the most recent 4 commits and added 2 commits: https://github.com/E3SM-Project/zstash/pull/402/commits/137c101f69566fff30a06a76cae6ac50ef343164 removes extra write lines, https://github.com/E3SM-Project/zstash/pull/402/commits/976397c47364ebcd4573e7f6bbba72dace1af5c9 fixes tar extraction (this is the same change as the 4th commit that was removed).

The Chrysalis tests are passing. Remaining action items: 
- [ ] Test on Perlmutter (once HPSS is up)
- [ ] Visually inspect code changes
- [ ] Draft a code review guide

forsyth2:
Chrysalis tests pass as well.

zstash/extract.py
```diff
@@ -479,9 +471,9 @@ def extractFiles(  # noqa: C901
     keep_files: bool,
     keep_tars: Optional[bool],
     cache: str,
-    cur: sqlite3.Cursor,
     args: argparse.Namespace,
     multiprocess_worker: Optional[parallel.ExtractWorker] = None,
+    cur: Optional[sqlite3.Cursor] = None,
```
forsyth2: ## Database Connection Fix for Parallel Workers
**Function: `extractFiles()` - signature change**
- Swapped order of `multiprocess_worker` and `cur` parameters
- Made `cur` optional (defaults to `None`)
- **Why**: Each worker process needs its own database connection. SQLite connections can't be shared across processes. When `cur=None`, each worker opens its own connection and closes it when done.

forsyth2 (commented):
Here are explanatory comments from Claude re: the PR as these 5 diffs: 
```
> git log --oneline | head -n 6
0ebd7a5 Restore comments and type annotations
f33ea80 Claude's changes to reduce over-engineering
734f200 Fixes to make Perlmutter tests pass
a54c842 Implementation changes to support Python 3.14
e315362 Add Python 3.14 support
880b9fd Optimize zstash update by using scandir (#420)
```

All tests pass on Perlmutter. I still need to check on Chrysalis, but based on previous iterations, I don't expect issues there.

---

**Summary**: The main theme is replacing complex load-balancing + synchronization with simple, predictable ordering. This eliminates deadlocks at the cost of potentially less optimal work distribution (but for most workloads, round-robin is fine).

forsyth2:
Adding context for Copilot before marking it as a reviewer: The purpose of this pull request is to support Python 3.14. In order to make all tests pass, significant implementation changes were required to the parallelism workflow. The tests cases driving this change were `tests.integration.python_tests.group_by_command.test_extract_parallel.TestExtractParallel.testExtractParallel` and `tests.integration.python_tests.group_by_command.test_extract_parallel.TestExtractParallel.testExtractParallelHPSS`. (Note that other tests failed as well, but fixes to get those passing were minimal).

All tests pass on the code as presented in this pull request now.

Things to check: 
- Is this implementation change really necessary to support Python 3.14 while keeping tests passing?
- Is the new implementation efficient enough for transferring terabytes of data? I.e., is replacing the heap method with round-robin a mistake?

forsyth2:
@tomvothecoder How did you get Copilot to author https://github.com/E3SM-Project/simboard/pull/105? Does it need to be triggered from an issue specifically? I don't see an issue linked to that PR. Is my [request above](https://github.com/E3SM-Project/zstash/pull/402#issuecomment-3937841508) in the wrong place? Too complex? All I got from Copilot was the eyes emoji.

I see @chengzhuzhang had Copilot start https://github.com/E3SM-Project/zppy/pull/779 (it looks this was done by marking it as an assignee on https://github.com/E3SM-Project/zppy/issues/756) and https://github.com/E3SM-Project/zppy/pull/780 (again, looks to be done by assignment on https://github.com/E3SM-Project/zppy/issues/626)

forsyth2:
I rebased to include [a fix](https://github.com/E3SM-Project/zstash/pull/402/commits/562c67c8c6aadeb72e1637e8e1adee97ed0e5ace) specifically linked to the actual DevOps work and not the implementation changes

forsyth2:
Note that a full explainer on parallelism in `zstash` as it exists on the `main` branch can be found at https://github.com/E3SM-Project/zstash/discussions/426

zstash/extract.py
```diff
@@ -335,46 +333,52 @@ def multiprocess_extract(
     # A min heap, of (work, worker_idx) tuples, work is the size of data
     # that worker_idx needs to work on.
     # We can efficiently get the worker with the least amount of work.
-    work_to_workers: List[Tuple[int, int]] = [(0, i) for i in range(num_workers)]
-    heapq.heapify(workers_to_tars)
+    work_to_workers: List[Tuple[float, int]] = [(0.0, i) for i in range(num_workers)]
+    heapq.heapify(work_to_workers)
```
forsyth2: This was apparently a mistake that was fixed in this PR (see #426). That is, it should have always been `work_to_workers` here, despite that line being added over 6 years ago.

forsyth2:
@copilot Please address all the comments from the code review agent unless I've specifically noted it as an unnecessary change.

Issue #374, opened 2025-06-05
[Bug]: tar files are not deleted after successful globus transfer


tangq:
### What happened?

The documentation indicates that "--keep to keep a copy of the tar files on the local file system after they have been transferred to HPSS. Normally, they are deleted after successful transfer." So, without `--keep`, the tar files should be deleted after successful transfer. But the tar files are not deleted when transferred successfully via globus. I might not use the zstash command correctly. If so, sorry for the false alarm.

I reproduced this with a simple test on Chrysalis.
`touch tst.txt`
`zstash create -v --non-blocking --hpss=globus://nersc/~/E3SMv3_dev/tst --maxsize 128 .`

The log line `Calling hpss_put to dispatch archive file 000000.tar [keep, non_blocking] = [False, True]` confirms the keep flag is `False`.

Here is the complete log:
```
For help, please see https://e3sm-project.github.io/zstash. Ask questions at https://github.com/E3SM-Project/zstash/discussions/categories/q-a.
DEBUG: 20250605_035238_623132: Running zstash create
DEBUG: Local path : /lcrc/group/e3sm2/ac.qtang/tst
DEBUG: HPSS path  : globus://nersc/~/E3SMv3_dev/tst
DEBUG: Max size  : 137438953472
DEBUG: Keep local tar files  : False
DEBUG: Making sure input path exists and is a directory
DEBUG: 20250605_035238_623404:Calling globus_activate(hpss)
INFO: Creating client of type <class 'globus_sdk.services.auth.client.native_client.NativeAppAuthClient'> for service "auth"
INFO: Finished initializing AuthLoginClient. client_id='6c1629cf-446c-49e7-af95-323c6412397f', type(authorizer)=<class 'globus_sdk.authorizers.base.NullAuthorizer'>
INFO: Setting up RefreshTokenAuthorizer with auth_client=[instance:23455989535200]
INFO: Setting up a RenewingAuthorizer. It will use an auth type of Bearer and can handle 401s.
INFO: RenewingAuthorizer will start by using access_token with hash "6e0977482bceb074128f911539d014c5907e30d53bcf7fa9c0c1d1b7240ff704"
INFO: Setting up an AccessTokenAuthorizer. It will use an auth type of Bearer and cannot handle 401s.
INFO: Creating client of type <class 'globus_sdk.services.transfer.client.TransferClient'> for service "transfer"
[...]

PR #404, opened 2025-11-24
Add test for tar deletion

forsyth2:
## Summary

Objectives:
- Add test for tar deletion. It should fail on `main` when using `--non-blocking`, given #374. After merging #405, it should pass. (For this reason, I specifically made this a separate pull request, allowing us to test both on `main` _and_ with the fixes of #405).

Issue resolution:
- This adds an explicit test for #374.
- This will allow us to check if #405 solves the issue.

## Small Change

- [x] To merge, I will use "Squash and merge". That is, this change should be a single commit.

### DIFF
--- a/tests/README.md
+++ b/tests/README.md
@@ -6,6 +6,7 @@
 tests/
   integration/ # Tests that call zstash from the command line
     bash_tests/ # Test zstash commands via bash scripts
+      run_from_any/ # Run these from any machine
       run_from_chrysalis/ # Run these from Chrysalis (these use Globus and/or require the Chrysalis file system)
       run_from_perlmutter/ # Run these from Perlmutter (these use `hsi` directly and/or require the Perlmutter file system)
     python_tests/ # Test zstash commands via Python unittest wrappers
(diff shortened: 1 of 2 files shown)


tests/integration/bash_tests/run_from_chrysalis/test_globus_tar_deletion.bash
```diff
@@ -0,0 +1,170 @@
...
+
+    echo "Checking src"
+    ls ${src_dir}/${case_name}/ 2>&1 | tee ls_${case_name}_src_output.log
+    check_log_has "${case_name}.log" ls_${case_name}_src_output.log
+    check_log_has "zstash_demo" ls_${case_name}_src_output.log
+    ls ${src_dir}/${case_name}/zstash_demo 2>&1 | tee ls_${case_name}_src2_output.log
+    check_log_has "zstash" ls_${case_name}_src2_output.log
+    ls ${src_dir}/${case_name}/zstash_demo/zstash 2>&1 | tee ls_${case_name}_src3_output.log
+    check_log_has "index.db" ls_${case_name}_src3_output.log
+    check_log_does_not_have "tar" ls_${case_name}_src3_output.log # tar is deleted at src
```
forsyth2: Passes for blocking, fails for non-blocking, confirming #374.

chengzhuzhang:
@forsyth2 I I think the change looks okay. my preference would be having this type of specific testing change together with the actual code change to reduce the number of PR and review cycle.

Issue #249, opened 2023-01-21
zstash is impractical for exascale simulations


PeterCaldwell:
`zstash create` is way too slow to be practical for km-scale global simulations. The issue seems to be collecting all the files into a tar file before moving it to HPSS. I think zstash also uses hsi under the hood and OLCF says hsi is much slower than htar, which is much slower than globus. I ended up having to use a 1 node batch job to try to do this. It ran for the max allowable 2 days then timed out, so I restarted it with `zstash update` and had to resubmit it every few days for weeks. I think one of my cases might have finished, but I never was able to verify if all the data got uploaded. 

`zstash check` is also impractical for these big simulations - it takes weeks to upload the data and even longer to download it again. Perhaps there could be a "quickcheck" option that just ensures that all the files in the original directory are found in the zstash index and that the total file size is consistent with the sum of the individual files you tarred? I'm also still a bit confused why we can't checksum the data on HPSS without moving it back to scratch. 

All of this isn't to say that zstash isn't great for what it was made for and it's fine if we never end up using it for exascale simulations. I'm just providing my experience as the first exascale zstash user.

chengzhuzhang:
Hi @PeterCaldwell Thank you for checking out the tool. I think it is the first application of zstash on exascale simulations, and zstash archiving is not practiced too often on OLCF. Since you also mentioned Globus, I'm wondering what is your workflow using zstash (i.g., archive from OLCF disk to HPSS, OLCF disk to NERSC HPSS, etc). I think it would be helpful to include the zstash command line here (with the data paths), so that developers can understand scale of the data and help diagnose/identify the bottle neck, and to try make zstash useful for exascale simulations!

golaz:
@PeterCaldwell : thanks for trying out and sharing your experience, as well as detailing your eventual workflow:

https://github.com/E3SM-Project/scream/issues/2131

Initial action items from your experience:

1. Implement the `--include` functionality, see https://github.com/E3SM-Project/zstash/issues/199
2. Understand the reason for the speed difference between direct transfer to HPSS via globus compared to what zstash is doing. If the local tar file generation is the bottleneck, we could consider bypassing it and directly archive to HPSS for large files.

PR #427, opened 2026-02-26
Add performance profiling as a standard zstash test

forsyth2:
## Summary

#414 introduced some early performance profiling for `zstash`, but these changes were not ultimately included in `main`. However, performance is becoming an ever more important feature of `zstash`, as that work showed. Other issues relating to performance: #402/#424, #249.

Objectives:
- Test `zstash` runtimes on various combinations of variables:
  - `create`, `update`, `extract`
  - `extract` sequential vs parallel mode (especially relevant for #402/#424)
  - `--hpss` as "none", an HPSS path, or a Globus path

## Small Change

### DIFF
--- a/conda/dev.yml
+++ b/conda/dev.yml
@@ -23,6 +23,9 @@ dependencies:
   # =======================
   - pytest
   - pytest-cov
+  - matplotlib-base
+  - pandas
+  - numpy
   # Documentation
   # =================
   # If versions are updated, also update in `.github/workflows/workflow.yml`
--- a/tests/performance/README.md
+++ b/tests/performance/README.md
@@ -1,0 +1,85 @@
+# How to profile zstash's performance
+
+Performance profiling should be done on Perlmutter. We're keeping the performance records in `/global/homes/f/forsyth/zstash_performance_records`.
+
+## Generate performance data.bash
+
+In `zstash/tests/performance/generate_performance_data.bash`, edit the run metadata:
+
+```bash
+# The performance data will end up in `results_csv="${work_dir}${unique_id}/results.csv"`
+# Use `/pscratch` since a lot of data will be transferred.
+# The results csv alone will be copied to a long-term (i.e., non-scratch) directory at the end.
+work_dir=/pscratch/sd/f/forsyth/zstash_performance/
+unique_id=performance_20260603
+
+# This is the environment that zstash will be run in.
+# Using Unified environment:
+environment_commands="source /global/common/software/e3sm/anaconda_envs/load_latest_e3sm_unified_pm-cpu.sh"
+# Example dev environment:
+environment_commands="source /global/homes/f/forsyth/miniforge3/etc/profile.d/conda.sh; conda activate zstash-pr427-20260603"
+```
+
+These parameters you probably won't have a need to change:
+```bash
+# These are the directories to run `zstash create`, `zstash update`, and `zstash extract` on.
+dir_to_copy_from=/global/cfs/cdirs/e3sm/forsyth/E3SMv2/v2.LR.historical_0201/
+subdir0=build/
+subdir1=run/
+subdir2=init/
+
+# This specifies which `--hpss` settings will be run:
+HPSS_OPTIONS=("none" "hpss" "globus")
+
+# This is what will be used for the "hpss" option:
+dst_hpss_path=/home/f/forsyth/zstash_performance
+
+# This is what will be used the "globus" option:
+fresh_globus=true # This will prompt a fresh Globus authentication
+dst_endpoint_uuid=15288284-7006-4041-ba1a-6b52501e49f1 # This is LCRC's endpoint
+dst_endpoint_archive_dir=/lcrc/group/e3sm/ac.forsyth2/zstash_performance_dst_dir/
+```
+
+Results will be saved to `${results_csv}` (recall `results_csv="${work_dir}${unique_id}/results.csv`). To keep all records together in a non-scratch space, the results csv is also copied to: `/global/homes/f/forsyth/zstash_performance_records/${unique_id}_results.csv`.
+
+## Visualize performance
+
+In `zstash/tests/performance/visualize_performance.py`, edit the run metadata:
+
+```python
+# The results to show in Fig. 1
+# This should be the results.csv you just generated in the step above.
+RESULTS_CSV: str = ""
+
+# The results to compare against in Fig. 2.
+# Set to None to skip Fig. 2.
+# This will typically be the second-to-oldest results.csv in the records space
+BASELINE_RESULTS_CSV: Optional[str] = ""
+
+# Output path for the saved figures.
+# Set to None to display interactively instead of saving.
+# Make sure to put this on the web server path,
+# i.e., /global/cfs/cdirs/e3sm/www/...
+OUTPUT_PATH: Optional[str] = ""
+```
+
+## For reference
+
+Records made before the long-term record space was made have been copied to it via:
+```bash
+SCRATCH_SPACE=/pscratch/sd/f/forsyth/zstash_performance
+RECORDS_SPACE=/global/homes/f/forsyth/zstash_performance_records
+
+for unique_id in \
+  performance_20260225 \
+  performance_20260226_pr402 \
+  performance_20260226_pr424 \
+  performance_20260226_pr428 \
+  performance_20260402 \
+  performance_20260414 \
+  performance_pr416_20260403 \
+  performance_pr416_20260406
+do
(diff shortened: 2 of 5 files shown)


forsyth2:
@chengzhuzhang

> There have been complaints about `zstash check` being slow.

I'm assuming you mean new/recent reports? I know we resolved the high-res team's issues by making more use of the `--tars` option.

> would it possible to do the performance profiling for zstash check ?

 I think that's good idea, but can we do it as a separate pull request? I think this PR is pretty close to merging, and I'd like to have _something_ on `main`. Furthermore, debugging `zstash check`'s slowness may give indicators of how best to profile it.

For what it's worth, Claude's assessment:

---

Looking at the code, adding `zstash check` would be a moderately sized addition — not trivial, but not a full rewrite either. Here's my assessment:

**Arguments for a separate PR:**

- The current PR already has a coherent, complete scope (create/update/extract profiling with full visualization). Shipping it now lets you get feedback and merge a working system before extending it.
- `check` has meaningfully different semantics — it validates archive integrity rather than measuring transfer/compression performance, so it may warrant its own visualization considerations (e.g., does it even make sense to compare `check` across HPSS modes the same way?).
- If something goes wrong with the `check` implementation, it won't block merging the core profiling work.

**Arguments for doing it now:**

- The plumbing is all there. Adding `check` to the bash script is straightforward: a `run_check()` function, a `record_result` call, and hooking it into the per-test loop after extract. The Python side just needs `check` added to `OP_ORDER`/`OP_TITLES`/`OP_DIR_COL` (or a new extract-style handler if it operates on the full archive).
- You'd avoid a second round of review setup, CI runs, and README updates for what is mechanically similar work.

**My recommendation: separate PR.** The current PR is clean and self-contained. `check` deserves a moment to think through what its x-axis and grouping should be — it runs on an existing archive, so it's more like `extract` than `create`, but it doesn't use workers in the same way. Rushing that design into this PR risks muddying both the code and the review. Merge what you have, then open a focused `check` PR once you've decided how it fits into the visualization layout.

tests/performance/generate/generate_performance_data.bash
```diff
@@ -0,0 +1,423 @@
...
+        record_result "$test_label" "$create_subdir" "$update_subdir" "$hpss_label" "create" "$create_log"
+
+        # --- UPDATE ---
+        run_update "$dir_to_copy_from" "$update_subdir" "$archive_dir" "$hpss_path" "$cache_dir" "$update_log"
+        record_result "$test_label" "$create_subdir" "$update_subdir" "$hpss_label" "update" "$update_log"
+
+        # --- EXTRACT (sequential=1 worker, parallel=2 workers) ---
+        for num_workers in 1 2; do
+            extract_log="${log_dir}extract_${hpss_label}_${num_workers}workers.log"
+            extract_dir="${mode_dir}extract_${num_workers}workers/"
```
forsyth2, line 458: Just a note for implementing the performance profiling piece of #446 

For this PR, we run `extract` twice on the same `create-update` sequence. This can be done simply by using a different extraction dir each time. So, for profiling `check`, we just need a third dir. Certainly no need to rerun the `create-update` sequence.

(There's still the question of how to fit `check` into the existing PDF, which is already fairly busy).

tests/performance/generate_performance_data.bash
```diff
@@ -0,0 +1,423 @@
...
+
+    # Create unique work directories for this test
+    work_subdir="${work_dir}${unique_id}/test${test_label}/"
+    mkdir -p "${work_subdir}"
+
+    log_dir="${work_subdir}logs/"
+    mkdir -p "${log_dir}"
+
+    # FIX: renamed `dst_hpss` local var to `dst_globus_path` so it does not
+    # shadow/overwrite the top-level `dst_hpss_path` parameter.
```
forsyth2: Can remove FIX lines

tests/performance/generate_performance_data.bash
```diff
@@ -0,0 +1,423 @@
...
+        run_update "$dir_to_copy_from" "$update_subdir" "$archive_dir" "$hpss_path" "$cache_dir" "$update_log"
+        record_result "$test_label" "$create_subdir" "$update_subdir" "$hpss_label" "update" "$update_log"
+
+        # --- EXTRACT (sequential=1 worker, parallel=2 workers) ---
+        for num_workers in 1 2; do
+            extract_log="${log_dir}extract_${hpss_label}_${num_workers}workers.log"
+            extract_dir="${mode_dir}extract_${num_workers}workers/"
+            mkdir -p "${extract_dir}"
+
+            # FIX: pass num_workers argument (was missing in original)
```
forsyth2: Can remove FIX lines

tests/performance/README.md
```diff
@@ -0,0 +1,85 @@
...
+# This is what will be used for the "hpss" option:
+dst_hpss_path=/home/f/forsyth/zstash_performance
+
+# This is what will be used the "globus" option:
+fresh_globus=true # This will prompt a fresh Globus authentication
+dst_endpoint_uuid=15288284-7006-4041-ba1a-6b52501e49f1 # This is LCRC's endpoint
+dst_endpoint_archive_dir=/lcrc/group/e3sm/ac.forsyth2/zstash_performance_dst_dir/
+```
+
+Results will be saved to `${results_csv}` (recall `results_csv="${work_dir}${unique_id}/results.csv`). To keep all records together in a non-scratch space, the results csv is also copied to: `/global/homes/f/forsyth/zstash_performance_records/${unique_id}_results.csv`.
```
forsyth2: Add a line about how to actually run each script. The current implementation only notes what metadata to change.

forsyth2 (commented):
In addition to [my latest test run](https://github.com/E3SM-Project/zstash/pull/427#issuecomment-4617702915), I've now done a high-level visual inspection of this largely-AI generated diff.

forsyth2 (commented):
I think this is ready to merge. Further additions are planned (notably #446 for `zstash check`, and maybe also some sort of averages plot built on many records over time).

@chengzhuzhang @TonyB9000  I imagine I'll be the primary user of this for-developers feature, but do either of you want to review it before I merge it?

conda/dev.yml
```diff
@@ -23,6 +23,9 @@ dependencies:
   # =======================
   - pytest
   - pytest-cov
+  - matplotlib-base
+  - pandas
+  - numpy
```
forsyth2: Note: In releases & RCs, we usually update dependencies required for the production environment. These 3 dependencies are being added to the "Testing" category and as such should not be added to the production environment.

For reference, https://github.com/E3SM-Project/zstash/pull/443 notes
```
# From https://github.com/conda-forge/zppy-interfaces-feedstock/pull/5#pullrequestreview-3351366258:
# Please always check the diff between the tag you're making a package for and the previous one. Please look for changes e.g. in dev.yml

# Review:
# https://github.com/E3SM-Project/zstash/compare/v1.6.0rc2...v1.6.0rc3
# No conda/dev.yml changes to include in recipe/meta.yaml.
```

tests/performance/generate/developer_run.cfg
```diff
@@ -0,0 +1,19 @@
...
+subdir2=init/
+
+HPSS_OPTIONS=none hpss globus
+dst_hpss_path=/home/f/forsyth/zstash_performance
+
+fresh_globus=true
+dst_endpoint_uuid=15288284-7006-4041-ba1a-6b52501e49f1
+dst_endpoint_archive_dir=/lcrc/group/e3sm/ac.forsyth2/zstash_performance_dst_dir/
+
+performance_archive_dir=/global/homes/f/forsyth/zstash_performance_records
```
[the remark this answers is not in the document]
forsyth2: This is specifically for me to have a ready-to-go cfg. The blank slate cfg is `perf.cfg`.

tests/performance/visualize/developer_run.cfg
```diff
@@ -0,0 +1,14 @@
...
+results_csv=/pscratch/sd/f/forsyth/zstash_performance/performance_20260603/results.csv
+baseline_results_csv=/pscratch/sd/f/forsyth/zstash_performance/performance_20260414/results.csv
+output_path=/global/cfs/cdirs/e3sm/www/forsyth/zstash_performance/performance_pr427_20260603.png
+
+viz_run_id=viz_test_20260618_try2
+most_recent_gen_run_id=performance_20260603
+
+hpss_filter=none,hpss,globus
+
+figures=
```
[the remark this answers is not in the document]
forsyth2: This is specifically for me to have a ready-to-go cfg. The blank slate cfg is `perf.cfg`.

